PR #14, opened 2018-08-23
12 Feature Content Walls

### LINKED ISSUE #12: Content Wall
The content wall would add the ability to implement a pay wall, age gate, terms-of-use acceptance, etc. before being able to access a page.

The content wall would be a somewhat generic functionality, probably either in the settings or at the page level. Contents of the wall could be customized with full `LAYOUT_STREAMBLOCKS` or `CONTENT_STREAMBLOCKS`. We might want to also add the ability to have a light weight form (newsletter sign up, yes/no buttons, checkbox, etc.). Or alternatively we could let the user implement that themselves in HTML.

Questions:
* Should this be at the page instance level? With option to apply to all sub-pages or just this page?
* Should it be a snippet in similar vein to the navbar, and have the option within the snippet to specify which pages this applies to? (This would enable the user to create multiple walls for different content)
* Or a setting? We would probably still need some kind of clusterable sub-model to enable multiple walls.

Default built-in template could be a modal, with option to make the modal dismissible or non-dismissible.

corysutyak:
Fixes #12


coderedcms/models/snippet_models.py
```diff
@@ -211,7 +211,36 @@ class Meta:
     def __str__(self):
         return self.name
 
+@register_snippet
+class ContentWall(models.Model):
+    """
+    Snippet that restricts access to a page with a modal.
+    """
```
vsalvino: Should probably add a custom CSS class and ID that both get put in at the top-level div of the modal, so that the developer can easily target them with css/js if needed.

PR #18, opened 2018-09-26
Feature organization

corysutyak:
Reorganizes the project tree structure.

Instead of everything living in the `coderedcms` app, the `coderedcms` app is divided into sub apps.  Right now this means that there is a `core` and a `search` sub app in the `coderedcms` app.  This was done to better mirror Wagtail's project structure, add better modularity and better testing organization.  

The downside is that this will break existing sites using the cms.


vsalvino:
I am OK with the new structure proposed in this PR, however the migrations must seamlessly handle the change without breakage for existing projects.

This change should also come with a documentation update to explain how and when to organize new features into core, or a separate module.

DOCS.md
```diff
@@ -48,7 +48,7 @@ Similar to Navbars, footers are also managed as snippets and also render top dow
...
 
 ### Custom templates
-The templates directory inside the `website` app is empty by default. Any templates you put in here will override the default coderedcms templates if they follow the same name and diretory structure. This uses the standard Django template rendering engine. For example, to change the formatting of the article page, copy `coderedcms/templates/coderedcms/pages/article_page.html` to `website/templates/coderedcms/pages/article_page.html` and modify it.
+The templates directory inside the `website` app is empty by default. Any templates you put in here will override the default coderedcms templates if they follow the same name and diretory structure. This uses the standard Django template rendering engine. For example, to change the formatting of the article page, copy `core/templates/core/pages/article_page.html` to `website/templates/core/pages/article_page.html` and modify it.
```
vsalvino: General comment that applies across the whole pull request. I do not like that "coderedcms" is no longer part of the template path. "core" is such a generic word that it could easily conflict with other pip packages. Also, from the client project perspective, it is not very clear that templates in "core" are part of coderedcms.

I do like the current naming convention, because when looking at a project that uses coderedcms, it is readily apparent that templates in the coderedcms directory are overriding the defaults. If anything, I would propose keeping it as-is or updating to "coderedcms/core/"

vsalvino:
We decided not to go down this route at this time.

PR #27, opened 2018-11-13
Feature LocationPage and LocationIndexPage

### LINKED ISSUE #26: LocationPage and LocationIndexPage
We need to develop a concept of a `LocationPage` and a `LocationIndexPage`.  The idea here would allow certain websites (ie, restaurants, retailers, etc) to show a list of their locations that are searchable.  This would be done with the `LocationPage`.  The `LocationPage`s would live under a `LocationIndexPage`.  The `LocationIndexPage` would show a google map with the locations of all the `LocationPage`s.  It would then be able to be searched upon.

corysutyak:
Fixes #26 

`LocationPage`
* An extension of `CoderedWebPage`
* Holds `address`, `latitude`, and `longitude` information
* User is meant to provide an address, unless specified otherwise, the CMS will calculate the `latitude` and `longitude` with `geocoder`.  While I still strongly feel that using `django-address` to handle addresses is a much better path, it is not feasible until it won't break at runtime for not having your api key stored in your settings file.
* The template for the page itself has a stripped down `body` field that just holds the `CONTENT_BLOCKS`.  It will also output the address and a map (toggleable) showing the location.
* Settings exist for the page to determine the content of the page's map marker.

`LocationIndexPage`
* Implements a non-trivial google map that will dynamically load markers based on the bounds of the map's viewport.
* Depending on specified page settings, you can also view a searchbox and a list of markers that are in the map's viewport.' Settings also exist to change the map's default zoom, lat, and lng.
* Because it is an `IndexPage` it holds the ability to list it's children pages.  This will probably be an unused feature for most cases.

Misc
* Abstracted the `ArticleIndexPage` template to allow `LocationIndexPage` to leverage it.
* Map marker data is obtained by making an ajax call to the page he map is being loaded on.  The pages that are going to load map marker data have an ajax view to handle response data.  The response data comes in the form of the GeoJSON spec. (http://geojson.org/). 
* `IntegrationSettings` has been added as a wagtail setting.  Currently it just has a variable to hold the Google API Key for the Maps.


coderedcms/models/page_models.py
```diff
@@ -1053,3 +1056,239 @@ def serve_preview(self, request, mode):
...
+    )
+    longitude = models.FloatField(
+        blank=True,
+        null=True,
+        verbose_name=_("Longitude")
+    )
+    auto_update_latlng = models.BooleanField(
+        default=True,
+        verbose_name=_("Auto Update Latitude and Longitude"),
+        help_text=_("If checked, automatically update the latitude and longitude for Google Maps when the address is updated.")
```
vsalvino: Remove "google" verbiage.

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1056,239 @@ def serve_preview(self, request, mode):
...
+    )
+    show_map = models.BooleanField(
+        default=True,
+        help_text=_("If checked, this will show a Google Map with this location."),
+        verbose_name=_("Show Google Map")
```
vsalvino: Remove google verbiage

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1056,239 @@ def serve_preview(self, request, mode):
...
+        blank=True,
+        max_length=255,
+        verbose_name=_("Map Title"),
+        help_text=_("If this is filled out, this is the title that will be used on Google Maps.")
+    )
+    map_description = models.CharField(
+        blank=True,
+        max_length=255,
+        verbose_name=_("Map Description"),
+        help_text=_("If this is filled out, this is the description that will be used on Google Maps.")
```
vsalvino: Remove google verbiage

coderedcms/forms.py
```diff
@@ -6,15 +6,15 @@
...
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.http import HttpResponse
 from django.utils.translation import ugettext_lazy as _
-from django.shortcuts import render
 from wagtail.contrib.forms.views import SubmissionsListView as WagtailSubmissionsListView
 from wagtail.contrib.forms.forms import FormBuilder
 from wagtail.contrib.forms.models import AbstractFormField
 
+from coderedcms import utils
```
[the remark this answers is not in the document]
corysutyak: It's a pylint detecting error.

coderedcms/templates/coderedcms/pages/location_page.html
```diff
@@ -0,0 +1,37 @@
...
+            id="cr-map" 
+            data-zoom="14" 
+            data-latitude="{{ page.latitude }}" 
+            data-longitude="{{ page.longitude }}" 
+            data-geojson-url="{{ request.path }}?data-format=geojson"
+            data-key="{{ google_api_key }}"
+            data-callback="initMap"
+            data-libraries="places"
+            data-show-list="False"
+            data-show-search="False">
```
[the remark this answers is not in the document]
corysutyak: Using the iframe would prevent markers from working, as well as centering the map on a user's location.  The iframe also doesn't support lat/lng only place ids.

vsalvino:
Some additional comments from testing saucy staging:
***
Pin popup on map should have follwoing information:

Title
Address
[View Location](#)

***
Sidebar with locations should render below map on mobile. Also, these could possibly be limited with an `overlflow-y: scroll` css attribute or something to prevent it from getting too long when there are many results.
***
Sidebar location link should not use h5. Could probably use something like this to take up less space and provide more context:
```
<b>Title</b><br>
<small class="text-muted">Address</small>
```
***
Default location index page template should probably show page title and work with cover image (as similar as possible to the default page template). Hopefully this will also create a little space above the map so it is not butted right up against the previous element.
***
Change the "no locations" message to say something like: "There are no locations in this area. Try zooming out or moving the map to find a location."

WMeldon:
One suggestion I have is to add a few optional extra fields like website or phone number.  Also adding a way to link out to google maps itself would be nice for people on mobile.  That probably doesn't even need to be stored data, just generated from the long/lat already in the model.

corysutyak:
@vsalvino Is this something we want baked in or should that be something that's implemented on a per site basis?

README.md
```diff
@@ -93,7 +93,5 @@ CodeRed CMS is a pip package that essentially wraps Wagtail and provides marketi
 
 One major point of difference between between CodeRed and stock Wagtail is the approach to design and content. Wagtail being more of a CMS framework, is focused on a clear separation between design (UX) and content. We agree with this approach for larger informational sites. But as is usually the case with marketing sites, design and information are more tightly coupled. Developers shouldn’t *need* to create a new page type or a new block just to handle a design deviation that is used in one place on the site. Designers and editors shouldn’t *need* to engage the developer for every minor design-related change such as changing a CSS class. For this reason, CodeRed blurs the lines of design and content by enabling editors to specify templates on a per-page and per-block basis, CSS classes per-block, and many other logo, layout, and branding settings. We realize this is not the right approach for every site - but we do believe it adds a lot of value for marketing sites.
 
-
-
```
vsalvino: Exclude this file/change from the review. We have 3 lines between each section consistently.

coderedcms/models/page_models.py
```diff
@@ -545,6 +547,18 @@ def body_preview(self):
...
+    def page_ptr(self):
+        """
+        Overwrite of `page_ptr` to make it compatible with wagtailimportexport.
+        """
+        return self.base_page_ptr
+
+    @page_ptr.setter
+    def page_ptr(self, value):
+        self.base_page_ptr = value    
+
```
vsalvino: I'm assuming this override will not mysteriously break things in wagtail? Similar to the `get_children` that we previously had to quit overriding.

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1067,268 @@ def serve_preview(self, request, mode):
...
+    content_panels = (
+        CoderedWebPage.content_panels[:1] + 
+        [
+            FieldPanel('address'),
+        ] +
+        CoderedWebPage.content_panels[1:] +
+        [
+            FieldPanel('website'),
+            FieldPanel('phone_number'),
+        ]
```
vsalvino: These fields should probably be grouped together with address.

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1067,268 @@ def serve_preview(self, request, mode):
...
+            """
+            .format(
+                self.title,
+                self.address,
+                self.url
+            )
+        )
+
+    @property
+    def render_list_description(self):
```
vsalvino: Not a fan of generating HTML inside a model. Is this method really needed? Couldn't this just be part of the html template?

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1067,268 @@ def serve_preview(self, request, mode):
...
+
+    def save(self, *args, **kwargs):
+        if self.auto_update_latlng and GoogleApiSettings.for_site(Site.objects.get(is_default_site=True)).google_maps_api_key:
+            g = geocoder.google(self.address, key=GoogleApiSettings.for_site(Site.objects.get(is_default_site=True)).google_maps_api_key)
+            try:
+                self.latitude = g.latlng[0]
+                self.longitude = g.latlng[1]
+            except TypeError:
+                """Raised if google denied the request"""
+                pass
```
vsalvino: Shouldn't `geocoder.google()` go inside the try/except? i.e. in event of api timeout or someting. Also, I would think a message should be shown. Similar to wagtail's error/success messages on page save. Probably a "warning" level would do. E.g. "The address could not be geocoded".

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1067,268 @@ def serve_preview(self, request, mode):
...
+                    FieldPanel('center_longitude'),
+                    FieldPanel('zoom'),
+                ],
+                heading=_('Map Display')
+            ),
+        ]
+    )
+        
+    def geojson_data(self, viewport=None):
+        qs = self.get_index_children().live()
```
vsalvino: Add docstring including info about `viewport`, i.e. how it is pipe delimited and how the data is formatted.

coderedcms/project_template/website/models.py
```diff
@@ -80,3 +80,4 @@ class Meta:
         verbose_name = 'Web Page'
 
     template = 'coderedcms/pages/web_page.html'
+
```
vsalvino: Exclude/undo this change. File already has newline at the end, does not need another, makes the commit/PR look like it changes more files than it actually does.

coderedcms/admin_urls.py
```diff
@@ -1,10 +1,12 @@
...
+from wagtailimportexport import urls as wagtailimportexport_urls
 from wagtail.admin import urls as wagtailadmin_urls
-from coderedcms.views import clear_cache
-from coderedcms.settings import cr_settings
+from coderedcms.views import clear_cache, import_pages_from_csv_file
 
 
 urlpatterns = [
     url(r'^codered/clearcache$', clear_cache, name="clear_cache"),
+    url(r'^import-export/import_from_csv/$', import_pages_from_csv_file, name="import_from_csv"),
```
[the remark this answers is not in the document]
corysutyak: This follows the same url scheme that `wagtailimportexport` uses.  Is prefixing with codered still preferable?

coderedcms/models/page_models.py
corysutyak: It shouldn't.  If anything it'll help with compatibility with other wagtail packages.

coderedcms/models/page_models.py
corysutyak: This gets called in the template as well as the maps.js file.  If we only had to render it in the template I would agree.  But doing it this way prevents people from needing to rewrite the javascript if they want to change the layout.

coderedcms/views.py
```diff
@@ -125,3 +135,120 @@ def robots(request):
...
+        page.url_path = None
+        page.content_type = page_content_type
+        parent_page.add_child(instance=page)
+
+        # Custom Code to add the new pk back into the original page record.
+        page_record['content']['pk'] = page.pk
+
+        pages_by_original_id[page.id] = page
+
+    for (i, page_record) in enumerate(import_data['pages']):
```
vsalvino: Seems like this could be optimized to not have to loop through pages twice.

coderedcms/models/page_models.py
corysutyak: An error won't be raised if it times out.  The geocoder would just return an object with timed-out request.  I'm not entirely sure if we are able to add a message in this context, but I will try.

coderedcms/views.py
```diff
@@ -125,3 +135,120 @@ def robots(request):
         {'robots': robots},
         content_type='text/plain'
     )
+
+@transaction.atomic()
+def import_pages(import_data, parent_page):
```
vsalvino: For all of these utility functions and class, would it make more sense to put those in a separate file, maybe `importexport.py` or `utils.py` or something? So as not to clutter the actual views. I would picture only the actual view function(request) being here.

setup.py
```diff
@@ -39,6 +39,8 @@
         'pygments>=2.2.0,<3.0',
         'wagtail==2.2.*',
         'wagtailfontawesome>=1.1.3,<2.0',
+        'geocoder==1.38.1',
```
vsalvino: Would it be safe to use <2.0?

setup.py
```diff
@@ -39,6 +39,8 @@
...
+        'wagtail-import-export==0.1'
```
vsalvino: Would it be safe to use `<0.2.0` ?

coderedcms/admin_urls.py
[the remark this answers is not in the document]
vsalvino: Well, if the `import_from_csv` is our own view, then I would prefix with codered. If it is their view, then keep as-is. We might want to try and contribute these changes back to the import-export pip package in the future if it is generic enough.

coderedcms/models/page_models.py
vsalvino: Oh, ok. I don't know much about the page_ptr. Just wanted to double check.

coderedcms/models/page_models.py
vsalvino: Hmm... interesting. Could it get extracted into an {% include %} then, to get rendered in both places?

coderedcms/views.py
corysutyak: For simple cases yes, but it would probably break on complex page imports if we didn't lay out the structure of base pages first.  That was wagtail's rationale for doing it this way at least.

coderedcms/models/page_models.py
vsalvino: Ok. I would still feel better putting in in the try block, for any future unforeseen breakage, especially since it is a 3rd party package. To prevent people from not being able to save their pages if it happens to break.

setup.py
corysutyak: That's for us to decide I suppose.  I'm not sure what our dependency policy is.

coderedcms/admin_urls.py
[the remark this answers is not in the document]
corysutyak: It's our own view that copies much of what wagtail does with their importing.  I had to change a few things to match the use case of our csv's.  I don't think it would be generic enough for that package.  Their idea is that you're doing whole/partial tree dumps/uploads.  Not giving it a csv and telling it you want these entries as pages.  I can see the argument for using our url from that perspective, I just wasn't sure if it was correct.

coderedcms/models/page_models.py
```diff
@@ -1053,3 +1069,247 @@ def serve_preview(self, request, mode):
...
+    @property
+    def render_list_description(self):
+        return render_to_string(
+            'coderedcms/includes/map_list_description.html',
+            {
+                'url': self.url,
+                'name': self.geojson_name,
+                'address': self.address,
+                'description': self.geojson_description
+            }
```
vsalvino: One tiny change... please pass in `self` rather than the individual attributes, so devs would be able to customize the map_list_description template with whatever data they want.

coderedcms/views.py
```diff
@@ -125,3 +127,36 @@ def robots(request):
         {'robots': robots},
         content_type='text/plain'
     )
+
+
+def import_pages_from_csv_file(request):
```
vsalvino: Probably should have `@login_required` if this is an admin functionality?

setup.py
vsalvino: Go with `>=1.38.1,<2.0`

setup.py
vsalvino: Go with `>=0.1,<0.2`

PR #33, opened 2018-11-29
Support Wagtail 2.3 and Django 2.1

FlipperPA:
* Support the latest Wagtail, 2.3
* Support the latest Django, 2.1
* Upgrade `urls.py` files to use `path`, `re_path` instead of `url` (I *think* I got them all!)
* Some PEP-8 fixes.

I've tested this with a test project (`coderedcms start mysite`) and it migrates and brings up the site and admin.


vsalvino:
My testing checks out, I think we are good to go. I did have to make one commit with a couple minor CSS changes due to wagtail 2.3 style changes. Not sure of the proper workflow on how to merge my commit into your pull request. But once that is in this will be ready to merge.

https://github.com/coderedcorp/coderedcms/tree/FlipperPA-wagtail2.3-django2.1

Issue #9, opened 2018-08-10
Add class to the page title in it's html template.


corysutyak:
Currently there is no easy way to select a title of a page in the html with css.  We should make a change to add a class to the h1 for the titles to make them easily selectable.

vsalvino:
In a similar vain, I'm wondering if it would be a good idea to add custom predictable CSS class names to all of our default HTML templates. Similar to how wagtail generates CSS classes around blocks.

This would allow quicker customization without the need to create custom templates by coderedcms users.

vsalvino:
Could we consider this closed since the new `content_header` block was added to the web_page template per commit https://github.com/coderedcorp/coderedcms/commit/9b727b609d9d46d06cd0cabb70826d094cc0aa99 ?

vsalvino:
Closing this issue since it seems the page templates are extensible enough for now.

PR #40, opened 2018-12-07
Additional body blocks

vsalvino:
Adding additional body blocks in base.html. I think this would be more general than having them specifically in `web_page.html`. Also removing old unused reference in base.html.

This could probably close issue #9


coderedcms/templates/coderedcms/pages/web_page.html
```diff
@@ -1,32 +1,19 @@
...
-
-{% endblock %}
+    </div>
+  </div>
+  {% else %}
+  <div class="container my-5">
+    <h1>{{page.title}}</h1>
+  </div>
+  {% endif %}
+{% endblock %}
```
corysutyak: Missing newline

corysutyak:
newline is missing in html file, but looks good otherwise.

Issue #34, opened 2018-12-03
Show CodeRed CMS Version 


WMeldon:
Should probably start maintaining a version in the CMS that is visible on the admin panel and command line.  Most python projects appear to do this with a ```__version__``` variable in ```setup.py```

vsalvino:
Yes, I was thinking this exact same thing. Wagtail shows their version in the title attribute of the logo and under the settings panel in the admin. I propose that we replace this with our version; our version is pinned to a specific wagtail version it doesn't really need to show both. Thoughts?

WMeldon:
Sounds good to me.

PR #42, opened 2018-12-07
Show CodeRed CMS version in wagtail admin

vsalvino:
Fixes issue #34 .

Replaced wagtail version in settings panel with CodeRed CMS version. Also added the version on hover of the logo in the bottom left. Did not replace the wagtail version hover on the main brand image, as that would have required forking an entire template and didn't seem worth it.


corysutyak:
I like moving the version out of `setup.py`.  I would think it's necessary to fork the template for the wagtail version though.  It seems inconsistent to not show the same version everywhere.

vsalvino:
Your right... I might fork the wagtail template in a separate file, add a block, and then extend that template, so it will be easy to diff/update the forked template with future wagtail updates.

coderedcms/templates/wagtailadmin/shared/menu_settings_menu_item.html
```diff
@@ -0,0 +1,6 @@
+{% extends "wagtailadmin/shared/menu_settings_menu_item.html" %}
+{% load coderedcms_tags %}
+
+{% block menu_footer %}
+    <li class="footer"><div class="menu-item"><p class="wagtail-version">CodeRed&nbsp;CMS v&nbsp;{% coderedcms_version %}</p></div></li>
+{% endblock %}
```
corysutyak: Missing newline

vsalvino:
Ok should be good to go.

PR #45, opened 2018-12-09
Documentation

vsalvino:
Adding initial sphinx docs. Moved contents of DOCS.md into various docs as appropriate. Using a slightly customized classic sphinx theme, I think it looks nice and on-brand and should get the job done for now.

To do docs locally, install the cms in dev mode which will install additional development packages:
```
pip install -e ./[dev]
```
Then enter the docs/ directory and type `make html`. This will create output in `docs/_build/`. Just open up `docs/build/index.html` in your browser.

Check it out and let me know what you think. I put a build of the docs up at https://docs.codered.cloud/cms/. I was thinking we could use this as our own doc host, but probably under docs.coderedcorp.com for consistency.

I'm thinking under "Features" in the docs we should create a separate page for each major feature, and then maybe one page for all of the blocks. Structure of docs will have to evolve over time. But they are slightly more fun to write now that they look shiny :)


corysutyak:
It looks good and gets a pass from me.

I'm slightly hesitant on the custom css.  I don't normally see other sphinx implementation using their own css.  But it probably doesn't matter.

vsalvino:
Sphinx is extremely simple, just uses jinja templates. Their themes are designed to be used with custom css, they even have a basic theme that just outputs plain html and no css, which is designed for a starting point for custom themes.

My main concern on the custom look is that it looks good and matches our other coderedcms branding. What do you think of the design?

corysutyak:
It's a lot of red, which can hurt to stare at after a while, lol.  Swapping the blue and the red might be the way to go.

vsalvino:
Yea, I was thinking the same thing. I'll try out another design idea with the colors.

vsalvino:
Ok, I made some changes, am satisfied now. Check it out https://docs.coderedcorp.com/cms/

Issue #43, opened 2018-12-08
Remove formatted code block from default streamfield


vsalvino:
Topic of debate: the streamfield is extremely heavy when adding nested blocks. I'm thinking we could probably remove the Formatted Code block, since it is actually a bit if a pig due to the language selector dropdown, and is the least likely to be used by 99% of websites.

This would be something we would keep in the cms, just remove it from the default blocks on a web page. So it would give us another opportunity to figure out a clean and easy way for devs to add additional blocks to the streamfield without too much code.

I already have an idea, similar to how we have variables for CONTENT_BLOCKS etc, we could allow the user to override CONTENT_BLOCKS so they can "plug in" their desired blocks, and all the WebPages would automatically pull from the overridden CONTENT_BLOCKS (would still require a migration, but seems relatively painless otherwise). This would function similar to the CODERED_THEME_* settings.

FlipperPA:
If you head in this direction, we could use `wagtailcodeblock` as an option for folks. It uses PrismJS rather than Pygments, so it provides real-time syntax highlighting preview in the Wagtail editor interface. The set of languages given is set by a Django setting; if not set, it defaults to a typical set for Python programmers.

https://github.com/FlipperPA/wagtailcodeblock

vsalvino:
Let's completely remove the `CodeBlock` from coderedcms, and therefore remove pygments from the setup.py requirements. That should reduce a lot of overhead and migration problems.

Custom blocks can be added to the `body` of pages in `website/models.py` until we find a more dynamic approach of adding them in.

vsalvino:
Fixed by PR #50

PR #50, opened 2018-12-12
Remove CodeBlocks from CodeRed (WAS: Missing migration)

FlipperPA:
This migration is missing for the changes to the content field on several models downstream.


corysutyak:
So this is an issue we have that we haven't been able to nail down.  Whenever you make migrations, any of our concrete models that use our massive streamfield will create a new migration to alter the streamfield.  You can apply the migration, run `makemigrations` again and see that the exact same migration is created.  To our understanding, the created migration doesn't actually change anything in the database schema.  

More research needs to be done on finding out why this happens.  I suspect that at some point wagtail streamfields get sufficiently complex with all the nested blocks that `makemigrations` gets confused and doesn't think the schema is correct.  I have no evidence to support this thought, however.  It's also odd because I believe under the hood, streamfields are represented in the db as a string representation of a json object.  So you would think what you actually change in the streamfield doesn't matter as far as the schema is concerned.

What's even more bizarre is the fact that if I open up these migrations in sublime text, there is no syntax highlighting on the lines of code that contain the altered streamfield definition.

vsalvino:
Yes, in earlier experimentation we did manually set the migration to an empty streamfield, which worked perfectly fine. So in theory wagtail might be more ideal to generate a more generic migration for streamfields, rather than a streamfield with all of the blocks in the migration. But to Cory's point, Django will endlessly generate migrations for complex steamfields. This part of the streamfield code is way out of my depth, and I eventually gave up trying to change it for the purposes of coderedcms. Have any wagtail core members seen this behavior before? @tomdyson

FlipperPA:
@gasman might have some ideas too; I've run into this as well when a component of the streamfield changes. Could it be the list of languages built by Pygments?

gasman:
I haven't seen this issue myself, but I can offer some insights into what might be going on in the migrations framework, having hacked on that bit of Django in the past (https://code.djangoproject.com/ticket/24340 - the "rather esoteric use case" mentioned there was StreamField :-) )...

The logic for deciding "has this field definition changed or not" is centred on a method called `deep_deconstruct`, which converts a field definition into primitive python values (lists, tuples, strings etc) - if the values from the 'before' and 'after' state don't compare as equal, then the field is considered to have changed, and a migration gets created. If a field definition contains complex objects (such as StreamField blocks), they need special treatment within `deep_deconstruct` to ensure that they end up as simple values where the equality check behaves as expected. For StreamField blocks, this is done by providing [a `deconstruct` method](https://github.com/wagtail/wagtail/blob/5f7b7230a26e244fd9c39d98b6e206902b594a9e/wagtail/core/blocks/base.py#L331-L357).

If you're finding migrations are repeatedly being created when there's no actual change, it probably means that you've passed a complex object (i.e. one where two instances of the object created at different times do not compare as equal) as an argument to a block somewhere, and there's no special `deconstruct` treatment for it. To track that object down, your best bet is probably to progressively remove blocks from the StreamField definition until you find that repeated `makemigrations` commands no longer generate useless migrations.

As for why we generate migrations for things that have no effect on the database schema: as I understand it, this is a deliberate decision on Django's part, and while it often shows up in ways that seem useless (for example, generating a migration just because you fixed a typo in `help_text`...) there's a good reason for it. It ensures that when you access the 'frozen' version of the ORM (`apps.get_model('my_app', 'MyModel')`) for use in a data migration, you're getting back a version of the field that exactly matches the Python-level behaviour - not just database-level behaviour - as it existed at that time, and so any ORM methods used within the data migration should continue to work regardless of any changes in models.py. I believe that storing the complete StreamField block definition is in the spirit of that Django design choice.

vsalvino:
Thank you @gasman , this is extremely insightful.

The best approach for us might be to first remove FormattedCodeBlock (i.e. pygments) from the streamfield, and then empirically test various models and corresponding makemigrations to see what the output is. I suspect that pygments with its massive language selection dropdown is causing this problem for us.

@FlipperPA , I'm not going to act on this PR for the time being and prioritize issue #43 which should probably happen first before we get too deep into migration madness. You have helped further validate this elusive problem though, so thank you!

FlipperPA:
@vsalvino @corysutyak I may have figured this out. For a test, which I've pushed, I removed the CodeBlock and reran migrations; the first one ran, then it was okay:

```bash
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
Migrations for 'coderedcms':
  /vagrant/projects/coderedcms/coderedcms/migrations/0005_auto_20181214_1402.py
    - Alter field content on carouselslide
    - Alter field content on contentwall
    - Alter field content on footer
    - Alter field content on reusablecontent
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
No changes detected in app 'coderedcms'
```

I noticed in `setup.py` that Pygments is not pinned:

`pygments>=2.2.0,<3.0`

This likely means that the `get_all_lexers()` function is returning a separate list for each version of pygments, so whoever is running it may get a different list returned each time depending on the versino of pygments in the virtualenv. I was on version 2.3, so I rolled back to 2.1, and then 2.2, and then back to 2.3, and sure enough, a new migration each time:

```bash
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
Migrations for 'coderedcms':
  /vagrant/projects/coderedcms/coderedcms/migrations/0005_auto_20181214_1412.py
    - Alter field content on carouselslide
    - Alter field content on contentwall
    - Alter field content on footer
    - Alter field content on reusablecontent
(codered) [vagrant@vagrant codered]$ pip install pygments==2.1
Collecting pygments==2.1
  Downloading https://files.pythonhosted.org/packages/6c/9e/995fd69b4577a2011d610d3976d92551a0ffc642163ceacaa9e58977b28a/Pygments-2.1-py3-none-any.whl (754kB)
    100% |████████████████████████████████| 757kB 3.6MB/s
coderedcms 0.10.0 has requirement pygments<3.0,>=2.2.0, but you'll have pygments 2.1 which is incompatible.
Installing collected packages: pygments
  Found existing installation: Pygments 2.3.0
    Uninstalling Pygments-2.3.0:
      Successfully uninstalled Pygments-2.3.0
Successfully installed pygments-2.1
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
Migrations for 'coderedcms':
  /vagrant/projects/coderedcms/coderedcms/migrations/0006_auto_20181214_1413.py
    - Alter field content on carouselslide
    - Alter field content on contentwall
    - Alter field content on footer
    - Alter field content on reusablecontent
(codered) [vagrant@vagrant codered]$ pip install pygments==2.2
Collecting pygments==2.2
  Downloading https://files.pythonhosted.org/packages/02/ee/b6e02dc6529e82b75bb06823ff7d005b141037cb1416b10c6f00fc419dca/Pygments-2.2.0-py2.py3-none-any.whl (841kB)
    100% |████████████████████████████████| 849kB 6.1MB/s
Installing collected packages: pygments
  Found existing installation: Pygments 2.1
    Uninstalling Pygments-2.1:
      Successfully uninstalled Pygments-2.1
Successfully installed pygments-2.2.0
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
Migrations for 'coderedcms':
  /vagrant/projects/coderedcms/coderedcms/migrations/0007_auto_20181214_1414.py
    - Alter field content on carouselslide
    - Alter field content on contentwall
    - Alter field content on footer
    - Alter field content on reusablecontent
(codered) [vagrant@vagrant codered]$ pip install pygments==2.3
Collecting pygments==2.3
  Using cached https://files.pythonhosted.org/packages/fc/41/4f900a7852e25bb9350b4e3ee8c4aba0ee32abefd401456962b25f954823/Pygments-2.3.0-py2.py3-none-any.whl
Installing collected packages: pygments
  Found existing installation: Pygments 2.2.0
    Uninstalling Pygments-2.2.0:
      Successfully uninstalled Pygments-2.2.0
Successfully installed pygments-2.3.0
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
Migrations for 'coderedcms':
  /vagrant/projects/coderedcms/coderedcms/migrations/0008_auto_20181214_1414.py
    - Alter field content on carouselslide
    - Alter field content on contentwall
    - Alter field content on footer
    - Alter field content on reusablecontent
(codered) [vagrant@vagrant codered]$ ./manage.py makemigrations coderedcms
No changes detected in app 'coderedcms'
```

It looks like this might be what is going on. I'll issue a separate P.R. with Pygments pinned and a migration; let's see if that works for y'all? If so, this can be a stopgap on this problem until a decision can be made on `CodeBlock`. This is currently an issue for me as I'm integrating CodeRedCMS into an existing project, and cross-app migration dependencies are getting muddled with the lack of this one.

FlipperPA:
...and I think that should be a wrap. Second set of eyes appreciated, as always, @vsalvino!

vsalvino:
Looks, good. Please also remove ``coderedcms/blocks/code_block.html`` template and I will merge it in.

Since this was the migration thread, going to state here that I plan on re-generating all of new migrations as one migration before doing this release. So if anyone has been seriously working on a project from the master branch, your migrations will be broken after 0.10 is released.

vsalvino:
And FYI I have preserved the codeblock here https://github.com/vsalvino/coderedcms-codeblock

FlipperPA:
And done - file removed.

Migration bankruptcy is fun!

vsalvino:
Fixes issue #43

PR #52, opened 2018-12-14
Pin Pygments to 2.3.0 to fix tricky migration.

FlipperPA:
This PR pins Pygments to 2.3.0 to make the `get_all_lexers()` function return the same list. This stops the migration from being recreated over and over because the metadata the migration contains for the StreamField in the `CodeBlock` (hopefully).


FlipperPA:
Cool. I'll go back to the other PR to finish up.

PR #55, opened 2018-12-14
Preparing 0.10 release

vsalvino:
* Docs are responsive now.
* Doc updates.
* Regenerating migrations since 0.9.1 into one clean migration.

Please check out this branch and test it out with your existing projects to ensure a smooth upgrade to 0.10.


vsalvino:
Doing some initial testing, I can confirm there is no data loss going from "Embed Video" -> "Embed Media" (which are technically the same block with the same field name, so this was expected).

vsalvino:
Additional testing, can also confirm that a big component of migration issues is coming from use of `CODERED_FRONTEND_*` settings. Will update docs with warning to disparage usage of those for now.

Also discovered a simple bug in the modal block, it requires a footer and should not. Might try to roll that fix in to this release as well, since it will require a migration.

PR #60, opened 2018-12-18
Edited typo for pagelist_block.html

phyominh:
Found this typo while using **Latest Pages** from Content `StreamField`


corysutyak:
Thanks for spotting that!

Issue #44, opened 2018-12-08
Official theme package


vsalvino:
We already have the basis of a framework for 3rd party themes. But I think a good step would be actually trying to create an example 3rd party theme, documenting how this is done, and making whatever changes would be necessary in the cms to support this. The first example could be simple, still based on Bootstrap. It is possible to build a theme that swaps out bootstrap for something like foundation, materialize, skeleton, etc., but that is a task for a different day.

Requirements for the theme would be:
* Custom templates that override coderedcms templates
* Custom page and block templates in addition to coderedcms templates, utilizing the Layout or Adv Settings > Template chooser dropdown.
* Custom css and js added to base.html, to show how to properly do that.
* MAYBE some extra blocks or models that are part of the theme, to show how to integrate those into the cms as well.

The theme itself should be a separate pip package. But docs and any enhancements to support theming should be integrated into the CMS.

hayleyhartman:
**Beans**
this one seems pretty simple but definitely different from bootstrap and has been used to make a lot of different looking designs
[https://www.getbeans.io/showcase/](https://www.getbeans.io/showcase/)
[https://github.com/Getbeans/Beans](https://github.com/Getbeans/Beans
)

**Openstrap**
Looks significantly different from bootstrap in the nav bar area, seems to be pretty multi-purpose
[https://www.opencodez.com/free-themes/openstrap](https://www.opencodez.com/free-themes/openstrap)
[https://github.com/pavansolapure/openstrap](https://github.com/pavansolapure/openstrap
)

**StartBootstrap Creative**
This one is the most polished and modern looking I have seen so far and would probably require little restyling when we're implementing, I could also see a client wanting to implement this theme off the bat because it does look and function really nice.
[https://startbootstrap.com/themes/creative/](https://startbootstrap.com/themes/creative/)
[https://github.com/BlackrockDigital/startbootstrap-creative](https://github.com/BlackrockDigital/startbootstrap-creative)

**StartBootstrap Agency**
Also pretty polished and modern looking and looks like the components on there are significantly different from standard bootstrap components. I could also see clients wanting to use this theme out of the box, it looks really nice.
[https://startbootstrap.com/previews/agency/](https://startbootstrap.com/themes/agency/)
[https://github.com/BlackrockDigital/startbootstrap-agency](https://github.com/BlackrockDigital/startbootstrap-agency)

vsalvino:
[...]

PR #75, opened 2019-01-08
add custom style

jahss:


vsalvino:
Hi @jahss , thanks for contributing. This specific feature is something we have debated about adding. It's difficult to decide where the line should be drawn regarding this kind of customization in the admin (Adding CSS classes > Adding raw CSS styles > A full on CSS editor similar to WordPress, etc.).

So we're going to continue debating this before deciding whether or not to merge this in. Good job though, this PR looks pretty complete.

vsalvino:
Thanks for the effort in making this pull request. I think we will hold off on this feature for now. But we might reconsider in a future release once we begin work on themes per #44 .

PR #76, opened 2019-01-11
37 feature mailchimp integration

### LINKED ISSUE #37: Mailchimp integration
We need the ability to integrate form submissions with 3rd party APIs. Specifically for this use case, MailChimp.

### Scenario
Admin needs to be able to dynamically map which form fields will feed into MailChimp upon form submission. This does need to be dynamic as it will vary from form to form. Submissions will feed into specific MailChimp lists, groups, segments, or list sources on a per-form basis. Not sure on current set of MailChimp features, but at a minimum a list will be needed.

After the visitor submits the form on the website, the entry will be logged in coderedcms as usual, and the user will now be opted in to a mailchimp list.

### Implementation
Ideally some kind of dynamic API integration mechanism would be nice as part of the form page. But this might not be possible or be too difficult to implement. I'm fine with doing a mailchimp-specific feature. This could be an isolated component/package that uses a hook on form submit. Or it could be something baked into the form page. Not sure what is best, although I like the idea of using a hook.

From the technical standpoint, it needs to be fail safe so as not to bomb out or cause data loss if the API call fails. We still need to collect the form submission and notify an admin, rather than bubble this error up to the visitor. Perhaps building in some logging would be good for troubleshooting in different environments which could experience issues caused by IP whitlisting or domain verification.

From the admin UX standpoint, I envision this working similar to the CofirmationEmails, which uses the Django templating engine to map form submission fields within a context. That could be useful for rendering a JSON or POST payload to an API endpoint. Similar to ConfirmationEmails, maybe we have an ApiCall streamfield which takes a URL, an http method, and a payload to post (which gets rendered with the context).

corysutyak:
Fixes #37 

I still need to update the docs, but the code itself is ready for review.


coderedcms/models/page_models.py
```diff
@@ -420,17 +421,22 @@ def __init__(self, *args, **kwargs):
...
             self.index_show_subpages = self.index_show_subpages_default
 
+
     @classmethod
-    def get_edit_handler(cls):
-        """
-        Override to "lazy load" the panels overriden by subclasses.
-        """
-        return TabbedInterface([
+    def get_edit_handler_object_list(cls):
```
vsalvino: I'm just curious on the reason for breaking this up into 2 methods.

coderedcms/models/wagtailsettings_models.py
```diff
@@ -378,3 +378,19 @@ class Meta:
...
     )
+
+
+@register_setting(icon='fa-puzzle-piece')
+class MailchimpApiSettings(BaseSetting):
+    """
+    Settings for Mailchimp API services.
+    """
+    class Meta:
+        verbose_name = _('Mailchimp API Settings')
```
vsalvino: Can remove the word "settings" here (and also on google api settings) since they are nested under the "settings" menu panel.

vsalvino: Can remove the word "settings" here (and also on google api settings) since they are nested under the "settings" menu panel.

coderedcms/templates/coderedcms/formfields/mailchimp/subscriber_integration_js.html
```diff
@@ -0,0 +1,127 @@
+<script>
```
vsalvino: Ehh.. this js is a bit on the nasty side... It makes sense due to the dynamic nature of the integration, but we really need to find a better way to handle this kind of thing in the future. It's fine for now, but I will create an issue for all of the instances where we have to run JS through the django template engine. That can be an issue for Will maybe during pycon sprints.

coderedcms/api/mailchimp.py
```diff
@@ -0,0 +1,73 @@
...
+        self.access_token = MailchimpApiSettings.for_site(Site.objects.get(is_default_site=True)).mailchimp_api_key
+        if self.access_token:
+            self.set_base_url()
+            self.is_active = True
+        else:
+            self.is_active = False
+
+    def set_base_url(self):
+        """
+        The base url for the mailchimip api is dependaent on the api key.
```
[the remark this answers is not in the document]
corysutyak: Yup

https://developer.mailchimp.com/documentation/mailchimp/guides/get-started-with-mailchimp-api-3/#resources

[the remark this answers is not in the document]
vsalvino: I figured as much, just had to check.

coderedcms/models/page_models.py
corysutyak: It would be clearer when I have the documentation examples up.  But, it doesn't make sense for every `Page` type to have an integrations panel.  Breaking this up into two functions allows page types to append their own panels to the list of panels rather than rewriting the whole function for panels.

vsalvino: Ah I see. Another option might be to just add another panel, that only gets added when it's not empty. E.g.
``` python
if len(cls.integrations_panel) > 0:
    ObjectList(cls.integrations_panel, heading='Integrations', classname="integrations"), # classname corresponds to css class that could add the puzzle piece icon
```
Then on the concrete class:
``` python
integrations_panels = [ ... ]
```

Issue #39, opened 2018-12-07
Create login template


vsalvino:
When using wagtail's page privacy feature set to a fixed password, the login screen is a very basic unstyled design. We should probably provide a default bootstrap template that uses the site's styles and branding, or possibly extends `web_page.html`.

To reproduce:

* Create a page in the wagtail admin
* Select "Privacy"
* Set to require a password, and set a password.
* View the page. The default password input template is very plain. It should inherit from `pages/web_page_notitle.html` or `web_page.html` and use bootstrap styling for the form.

vsalvino:
Fixed in master

PR #78, opened 2019-01-11
Password required page template

kevincummings:
Fixes issue #39


coderedcms/templates/wagtailcore/password_required.html
```diff
@@ -0,0 +1,20 @@
+{% extends "coderedcms/pages/web_page_notitle.html" %}
+{% load i18n bootstrap4 %}
+
```
vsalvino: Question - does this page have a title that gets rendered in the browser tab? I'm assuming we probably should override the title block here.

coderedcms/templates/wagtailcore/password_required.html
```diff
@@ -0,0 +1,20 @@
...
+{% block content %}
+<div class="container">
+    <div class='text-center mt-5'>
+        <h3>{% trans "Password Required" %}</h3>
```
vsalvino: H1 would be more appropriate here.

coderedcms/templates/wagtailcore/password_required.html
```diff
@@ -0,0 +1,20 @@
...
+        {% block password_required_message %}
+            <p>{% trans "You need a password to access this page." %}</p>
+        {% endblock %}
+        <form action="{{ action_url }}" method="POST">
+            {% csrf_token %}
+            <div class="form-group">
+                {% bootstrap_form form layout='horizontal' %}
```
vsalvino: I think the `bootstrap_form` call should go outside of a `form-group` div... I believe in this case it would be generating nested `form-group` divs which is not recommended bootstrap syntax.

coderedcms/templates/wagtailcore/password_required.html
```diff
@@ -0,0 +1,20 @@
...
+                <input type="submit" value="{% trans 'Continue' %}" class="btn btn-primary" />
```
vsalvino: Please use `<button type="submit"` for better consistency with the rest of coderedcms. Make sure it still works with button instead of input.

PR #79, opened 2019-01-12
[REVIEW] Feature/docker

rosscdh:


Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN python setup.py install
+
+# uWSGI will listen on this port
+EXPOSE 8000
+
+# # Call collectstatic (customize the following line with the minimal environment variables needed for manage.py to run):
+# RUN DATABASE_URL=none /venv/bin/python manage.py collectstatic --noinput
+
+# Start uWSGI
+ENTRYPOINT ["coderedcms"]
```
vsalvino: Question... running the `coderedcms` command will not actually do anything by default. It is used as `coderedcms start` to create a new boilerplate project. So is this the most appropriate entrypoint command? I'm not really sure what the purpose of this docker image is.

rosscdh: yes this is true. I suspect i may have miudnerstood the purpose of this repo. but regardless an interesting project.

rosscdh: also using `coderedcms` as the entrypoint allows you to do things like `docker run --rm -v $PWD/clients/myclient:/src/output coderedcms:latest start`

rosscdh: ps yes it is the valid entrypoint; allows you to run start in your docker compose or kube config

rosscdh: `docker run --rm -v $PWD/clients/myclient:/src/output coderedcms:latest start my_project` which will output the generated to `/src/output`

Issue #61, opened 2018-12-18
CSS Issues in Wagtail Admin


FlipperPA:
The CSS included in for the Wagtail admin creates some weird horizontal spacing when integrated with an existing site. For example, have a look at the horizontal row (including Day of Week):

Without CodeRed (good):

![image](https://user-images.githubusercontent.com/68164/50177287-9c94e400-02cf-11e9-97b9-dea5ad684d65.png)

With CodeRed (bad - dropdowns obscured by the bar):

![image](https://user-images.githubusercontent.com/68164/50177438-01e8d500-02d0-11e9-91ea-85db1e60edf6.png)

The `content_panels` declaration is as follows; `FieldRowPanel` seems problematic with the CSS:

```python
    content_panels = Page.content_panels + [
        FieldPanel("group"),
        FieldRowPanel(
            [
                FieldPanel("day_of_week"),
                FieldPanel("start_time"),
                FieldPanel("end_time"),
            ]
        ),
        FieldPanel("types", widget=CheckboxSelectMultiple),
        FieldPanel("meeting_details"),
    ]
```

vsalvino:
Got it, yea that would be a bug.

Question... do you think it would be preferable to keep the Label above the Dropdown in the FieldRowPanel, or would it be preferable to have the Label next to the field as wagtail does it? I only ask because FieldRows tend to get a bit smashed/crowded when the inputs are not so short. I believe we had to do a lot of adjustments to labels and inputs because they would get way too smashed when streamfields start nesting (as in a typical `CoderedWebPage`)

vsalvino:
Snippet page also has a couple issues. Checkbox column looks off being so wide. Plus the buttons in the header look less than ideal stacked instead of side by site (not sure if there is actually not enough room due to font size, or not).

Specifically for the checkbox, I know there is CSS somewhere that normalizes table column widths, I can't remember why, but I do remember that was implemented because some table was out of whack due to customization.

Wagtail:
![image](https://user-images.githubusercontent.com/13453401/50200784-739f3e00-0325-11e9-91df-bc6c8248e412.png)

CodeRed:
![image](https://user-images.githubusercontent.com/13453401/50200766-62eec800-0325-11e9-9a49-f4416d17fd65.png)

FlipperPA:
[...]

PR #80, opened 2019-01-12
Removing some custom admin css to be more compatible with wagtail

vsalvino:
Should fix issues highlighted in #61 

Wagtail 2.4 started using a `.label-above` class, which this is tuned for, which greatly fixes the layout of the image editing page without us having to sacrifice streamfield label styles. These changes are compatible with both wagtail 2.3 and 2.4, but will provide more benefit to 2.4.

The main things that cause so much custom css at this point are:
1. using flexbox on labels/inputs so they render side-by-side nicely. Without this our deeply nested streamfields become ridiculously long and difficult to edit.
2. Making input boxes less padded and higher contrast. Once again, thick chunky inputs with an ultra-light font and super-light colors take up too much space and are difficult to read once they are nested in the streamfield.
3. Support the ability to actually scroll the side menu and settings menu when it overflows. This is something that is pretty badly broken in wagtail. This fix should probably be contributed upstream.

@FlipperPA I would love if you could test this branch out with your existing wagtail/coderedcms hybrid site and let me know if it looks better, or if this just makes it worse lol.

### DIFF
--- a/coderedcms/static/css/codered-admin.css
+++ b/coderedcms/static/css/codered-admin.css
@@ -88,7 +88,7 @@ input[type='checkbox']::before, input[type='radio']::before {
 
 @media screen and (min-width:50em) {
     body:not(.login) li.sequence-member .fields > li,
-    body:not(.login) ul.fields > li:not(.label-above) > .field {
+    body:not(.login) ul.fields > li:not(.label-above) .field {
         display: flex;
         flex-direction: row;
         flex-wrap: nowrap;
@@ -103,13 +103,16 @@ input[type='checkbox']::before, input[type='radio']::before {
         display:block;
         float:none;
         margin: 0;
+        width: auto;
+    }
+    body:not(.login) ul.fields > li:not(.label-above) .field > label {
         width:15em;
         min-width:15em;
     }
     body:not(.login) ul.fields > li:not(.label-above) ul.fields > li > label,
     body:not(.login) ul.fields > li:not(.label-above) ul.fields > li > .label,
-    body:not(.login) .multiple ul.fields > li:not(.label-above) .field > label,
-    body:not(.login) .field-content ul > li:not(.label-above) > label {
+    body:not(.login) .struct-block ul.fields > li:not(.label-above) .field > label,
+    body:not(.login) .multiple     ul.fields > li:not(.label-above) .field > label {
         width:10em;
         min-width:10em;
     }
--- a/coderedcms/static/css/codered-admin.css
+++ b/coderedcms/static/css/codered-admin.css
@@ -87,17 +87,17 @@ input[type='checkbox']::before, input[type='radio']::before {
 /* Display labels and fields horizontally, except .label-above */
 
 @media screen and (min-width:50em) {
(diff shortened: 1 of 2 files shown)


FlipperPA:
Sorry for the delay, @vsalvino - but this doesn't seem to have helped, same issue:

image

This is running against the `css-fix` branch with Wagtail 2.3. Wagtail 2.3 was still required in `setup.py` - I'll give it a test against Wagtail 2.4 now and report back.

FlipperPA:
The same issue exists with Wagtail 2.4. Here's the Page model:

```python
class Meeting(Page):
    """
    Model for storing meeting data.
    """

    DAY_OF_WEEK = (
        (0, "Sunday"),
        (1, "Monday"),
        (2, "Tuesday"),
        (3, "Wednesday"),
        (4, "Thursday"),
        (5, "Friday"),
        (6, "Saturday"),
    )

    STATUS_CHOICES = ((0, "Inactive"), (1, "Active"))

    group = models.ForeignKey(
        Group, null=True, blank=True, on_delete=models.SET_NULL, related_name="meetings"
    )
    meeting_location = models.ForeignKey(
        Location, null=True, on_delete=models.SET_NULL, related_name="meetings"
    )
    start_time = models.TimeField(null=True)
    end_time = models.TimeField(null=True)
    day_of_week = models.SmallIntegerField(default=0, choices=DAY_OF_WEEK)
    status = models.SmallIntegerField(default=1, choices=STATUS_CHOICES)
    meeting_details = models.TextField(
        help_text="Additional details for the meeting.", null=True, blank=True
    )
    types = ParentalManyToManyField(
        MeetingType,
        related_name="meetings",
        limit_choices_to={'intergroup_code__isnull': False},
    )

    @property
    def day_sort_order(self):
        """
        Returns 0 for today's day of the week, up to 6 for yesterday's day of the
        week rather than Sunday - Saturday.
        """
        day_sort_order = self.day_of_week - datetime.datetime.today().weekday() - 1
        if day_sort_order < 0:
            day_sort_order += 7

        return day_sort_order

    content_panels = Page.content_panels + [
        FieldPanel("group"),
        FieldRowPanel(
            [
                FieldPanel("day_of_week"),
                FieldPanel("start_time"),
                FieldPanel("end_time"),
            ]
        ),
        # HelpPanel(content=""),
        FieldPanel("types", widget=CheckboxSelectMultiple),
        FieldPanel("meeting_details"),
    ]
```

I've got a temporary work around with the blank `# HelpPanel(content=""),`. I'll take a deeper look when I get a chance, but I've been under the weather.

vsalvino:
@FlipperPA I was able to re-create your problem and fix. Issue was related to us setting a min-width on inputs. I found an alternate way to achieve the same thing that is more in line with wagtail's original way of sizing inputs.

@corysutyak this one should be ready for review and merge now. I have tested it with many possible scenarios.

vsalvino:
@FlipperPA fear not, 0.11 will be out soon and you can find more bugs :)

FlipperPA:
Just to confirm, the new master branch works beautifully! Thank you!

PR #105, opened 2019-01-31
102-fix-default-ordering

### LINKED ISSUE #102: Order child pages by default
Under **[edit page] > Layout > Order Child Pages By** there should be a dropdown option called "default" that will use the model's default order (which I believe would be `get_children`). Currently you have to select one of the options which does not respect Wagtail's manual ordering (or any other ordering set on the model class definition).

corysutyak:
Fixes #102 

This will respect manual ordering if `default` is selected.


coderedcms/models/page_models.py
```diff
@@ -133,8 +133,9 @@ class Meta:
 
     # Subclasses can override these fields to enable custom
     # ordering based on specific subpage fields.
-    index_order_by_default = '-first_published_at'
+    index_order_by_default = 'default'
```
vsalvino: I would prefer to see this be an empty string. Technically it would not work if there was a field called "default".

Issue #101, opened 2019-01-31
Search: ArticlePage results are returned in old->new order


vsalvino:
Ideally ArticlePage (or any other page type) that match the search query should be returned in the order specified by the parent page layout settings, or at a minimum newest to oldest as a better default.

corysutyak:
I'm not sure the best way to go about this.  We can surely organize the list of returned results by `date_published` or something similar.  But if the default search backend is organizing based on relevancy, any sorting after the fact will no longer respect that relevancy.  The same is true if results are organized by model type.  We would probably have to write our own search backend and force the CMS to only use that backend.  That looks non-trivial and we would be preventing users from doing something they normally can do with Wagtail.

corysutyak:
Fixed in PR #110

PR #110, opened 2019-02-13
Order default search model (CoderedPage) by last published date new->old

vsalvino:
Ordering is done on queryset before doing the search. The search backend is still free to reorder the results, however I don't believe the database search backend does any ordering.

When searching on a specific model, do not order, the assumption is the models' Meta ordering will be used.


corysutyak:
Fixes #101

PR #134, opened 2019-03-19
119 fix rich text block

### LINKED ISSUE #119: AMP errors in rich text block
It appears that wagtail has started adding custom attributes to the anchor `<a>` tag rendered from a rich text block, and these custom attributes are disallowed by AMP.

Specifically, a link to a page is rendering as:
```html
<a id="18" linktype="page">Page Name</a>
```
and according to google: "The attribute 'linktype' may not appear in tag 'a'."

The rich text block also renders images using the `img` tag. This should use `amp-img` when appropriate, similar to how the coderedcms image block template renders both formats.

corysutyak:
Fixes #119 

-Adds new template filter `richtext_amp` that will convert rendered html from a richtext field's shorthand representation into amp compliant html.


coderedcms/utils.py
```diff
@@ -37,3 +37,18 @@ def fix_ical_datetime_format(dt_str):
...
+def convert_to_amp(value):
+    """
+    Function that converts non-amp compliant html to valid amp html.
+    """
+    soup = BeautifulSoup(value)
+
+    #Replace img tags with amp-img
+    try:
+        img_tags = soup.find('img')
+        img_tags.name = 'amp-img'
```
vsalvino: One item of note, amp image tags must have a width and height specified - not sure how the richtext filter outputs them.

setup.py
```diff
@@ -48,7 +48,8 @@
         'wagtail==2.4.*',
         'wagtailfontawesome>=1.1.3,<2.0',
         'wagtail-cache==0.5.*',
-        'wagtail-import-export>=0.1,<0.2'
+        'wagtail-import-export>=0.1,<0.2',
+        'beautifulsoup4==4.4.0'
```
vsalvino: We can probably use same beautifulsoup version specifier as wagtail 2.4

coderedcms/utils.py
```diff
@@ -37,3 +37,18 @@ def fix_ical_datetime_format(dt_str):
         dt_str = dt_str[:-3] + dt_str[-2:]
         return dt_str
     return dt_str
+
+def convert_to_amp(value):
+    """
+    Function that converts non-amp compliant html to valid amp html.
+    """
```
vsalvino: Assuming value is a string? If so can you specify in the docstring.

coderedcms/blocks/__init__.py
```diff
@@ -16,7 +16,7 @@
 # Collections of blocks commonly used together.
 
 HTML_STREAMBLOCKS = [
-    ('text', blocks.RichTextBlock(icon='fa-file-text-o')),
+    ('text', blocks.RichTextBlock(icon='fa-file-text-o', template='coderedcms/blocks/rich_text_block.html')),
```
vsalvino: Can you confirm this won't trigger a migration? if so include the migration.

vsalvino: Actually, on second thought, what do you think about making our own richtext block (that just inherits from blocks.RichTextBlock, and specifies the template), so that anywhere it gets used will automatically have the template (and icon)?

corysutyak:
The `richtext` filter renders out the image height and width.  I added the migration (my bad), and made our own `RichTextBlock`

coderedcms/utils.py
corysutyak: The richtext filter outputs the width and height.

PR #141, opened 2019-03-25
67 feature enhance ga

### LINKED ISSUE #67: Enhance GA tracking
We can up our game at "not being evil" (which in Google speak means more tracking). When the **Settings > Tracking > Track all button clicks** setting is turned on, it does track all button clicks, but should be extended to track all link clicks, not just buttons. This includes links in the RichText  Editor, navbar, etc.

For a cleaner implementation, the tracking script currently in `button_block.html` could be reafctored into a global jquery event that fires on all `$('a').click()` events, and the tracking parameters could be provided in as data-attributes. Links without the advanced settings to specify tracking details should use reasonable defaults.

corysutyak:
Fixes #67 

Toggling the track all button settings will track all a tags and generate default categories/labels if they're not provided.


coderedcms/static/js/codered-front.js
```diff
@@ -216,4 +216,24 @@ $(document).ready(function()
             $(this).modal('show');
         }
     });
-});
\ No newline at end of file
+
+
+    /*** Tracking ***/
+    $('a').on('click', function(){
+        if($("meta[name='tracking:click']").length){
```
vsalvino: I would reverse the `if` and the `a.onclick` nesting so that the event is only wired up if we are going to track in the first place, otherwise every `click` would trigger an event that does nothing.

coderedcms/templates/coderedcms/pages/base.html
```diff
@@ -6,6 +6,9 @@
   <head>
         {% block tracking %}
         {% if settings.coderedcms.AnalyticsSettings.ga_tracking_id %}
+            {% if settings.coderedcms.AnalyticsSettings.ga_track_button_clicks %}
+                <meta name="tracking:click" value="True" />
```
vsalvino: I would probably replace this custom meta tag with a variable in the script below where google tag manager is pulled in:
```javascript
<script>
  tracking_click = true;
  window.dataLayer = window.dataLayer || [];
  ...
</script>
```
Then if the variable is undefined or false you know not to track the onclick event.

PR #170, opened 2019-04-30
issue 164

hayleyhartman:
Removed recurring empty div on rich text blocks by taking out the line that extends base-block to this div


hayleyhartman:
Issue 162 fixed by adding |urlencode to search url in google maps block. Issue 14 fixed by having rich-text block no longer extend base block

hayleyhartman:
changed this to only fix issue 164, see separate request for 162

vsalvino:
Looks like this still has both fixes in the same branch.

PR #171, opened 2019-05-01
Feature streamforms

corysutyak:
This is a first pass at an integration with https://github.com/noripyt/wagtail-flexible-forms and not currently ready to be merged.

### Big Changes:

- WFF is included as a copy in our code.  There code is largely left untouched apart from unregistering their FormSubmission modeladmin classes.
- Large refactor of `CoderedFormPage`.  The functionality of `CoderedFormPage` is essentially the same, but a lot of the functionality has been moved into a mixin.
- Implementation of `CoderedAdvancedFormPage`.  This page model integrates the WFF `StreamFormMixin` and our new `CoderedFormMixin`.  

### Little Changes:

- Changed the way `CoderedFormMixin` saves files to match how WFF saves files.  The only real difference is that they create a directory for the session key and then save the files in that directory.  It seems better that way.  It won't have any negative effects on previously uploaded files. 
- Added our own `CoderedFormAdmin` and `CoderedSubmissionAdmin` modeladmin classes.  WFF's versions of these classes work under the assumption that you're only using WFF and can break with Wagtail forms.

### TODO:

- [x] Integrate our StreamBlocks into the StreamForm

- [x] Add conditional logic StreamBlocks

- [x] Add Docs

- [ ] Test Mailchimp integration with StreamForms

### Things to Note:

- A handful of new migrations, one specific to WFF
- The `save_to_database` attribute for `CoderedFormMixin` is not respected by `CoderedAdvancedFormPage`.  Due to the form keeping state by saving an in progress submission based on the session, having the `save_to_database` attribute turn this feature on/off would require a bit of work on WFF.  I also question it's existence to begin with.  I'm not sure if it's something we need as I can't think of many good reasons why someone wouldn't want to save their form data.


corysutyak:
Integration with our blocks has been added.  The attributes for conditional logic has been added as well.  The conditional logic isn't hooked up yet though.

vsalvino:
One housekeeping item of note, please include the unmodified LICENSE file from the NoriPyt repository in the `wagtail_flexible_forms` directory.

coderedcms/models/page_models.py
```diff
@@ -1147,92 +1137,38 @@ def form_live(self):
...
-
-    def process_form_submission(self, request, form):
-        """
-        Accepts form instance with submitted data, user and page.
-        Creates submission instance.
-
-        You can override this method if you want to have custom creation logic.
-        For example, if you want to save reference to a user.
-        """
+    def process_data(self, form, request):
```
vsalvino: Please use docstring to document what this function does, and what the arguments are.

coderedcms/models/page_models.py
```diff
@@ -1147,92 +1137,38 @@ def form_live(self):
...
             else:
                 processed_data[key] = val
+        return processed_data
 
-        # Get submission
-        form_submission = self.get_submission_class()(
-            form_data=json.dumps(processed_data, cls=DjangoJSONEncoder),
-            page=self,
-        )
+    def get_storage(self):
```
vsalvino: Please add docstring to describe what this method does.

coderedcms/models/page_models.py
```diff
@@ -1291,24 +1227,22 @@ def process_form_submission(self, request, form):
...
         """
         Sends a form submission summary email.
         """
         addresses = [x.strip() for x in self.to_address.split(',')]
         content = []
-        for field in form:
-            value = processed_data[field.name]
-            # Convert lists into human readable comma separated strings.
+
+        for key, value in processed_data.items():
```
vsalvino: Did the data structure of `processed_data` change? Or are you just using python kung-foo to refactor it? Either way, please update the docstring to describe the method arguments.

coderedcms/models/page_models.py
```diff
@@ -1291,24 +1227,22 @@ def process_form_submission(self, request, form):
...
             if isinstance(value, list):
                 value = ', '.join(value)
             content.append('{0}: {1}'.format(
-                field.label,
+                key.replace('-', ' ').replace('_', ' ').title(),
```
vsalvino: It's not very DRY-ish that we are doing the string replace here, and also in the `data_to_dict` method. Is there a way to improve?

coderedcms/models/page_models.py
```diff
@@ -1291,24 +1227,22 @@ def process_form_submission(self, request, form):
...
                 utils.attempt_protected_media_value_conversion(request, value)
             ))
-        content = '\n\n'.join(content)
+
+        content = '\n'.join(content)
```
vsalvino: Please leave the double newline. The emails are not readable without it, because most forms have large textfields that start to run together when there is only one newline. Alternatively, it might actually help to "format" the plaintext by using a divider like `\n-------------\n` between the fields.

coderedcms/models/page_models.py
```diff
@@ -1394,18 +1431,61 @@ def serve(self, request, *args, **kwargs):
...
+                )
+                return self.serve_success(request, *args, **kwargs)
+            return HttpResponseRedirect(self.url)
+        return CoderedWebPage.serve(self, request, *args, **kwargs)
+
+    @classmethod
+    def get_submission_class(cls):
+        return CoderedSessionFormSubmission
+
+    def get_storage(self):
```
vsalvino: Isn't this already provided via the mixin?

coderedcms/project_template/project_name/settings/base.py
```diff
@@ -29,6 +29,7 @@
 
     # CodeRed CMS
     'coderedcms',
+    'coderedcms.wagtail_flexible_forms',
```
vsalvino: Is this really needed here? The code is already part of coderedcms. Unless there is some kind of wagtail magic where it needs to be its own app.

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -100,7 +103,15 @@ def process_form_cell(request, cell):
...
-    return localize(cell)
+
+    # This is a last resort rendering for links, used primarily for links stored in StreamForms.
+    # TODO, find a better way to handle this.
+
+    try:
+        unescaped_cell = unescape(cell)
+        return localize(mark_safe(unescaped_cell))
+    except:
+        return cell
```
vsalvino: I'm not entirely sure what's going on here. Does the streamform somehow store its own links to uploaded files?

coderedcms/blocks/__init__.py
```diff
@@ -5,49 +5,54 @@
 """
 
 from django.utils.translation import ugettext_lazy as _
+from wagtail.core.blocks import CharBlock, StreamBlock, StructBlock
 
+from coderedcms.wagtail_flexible_forms.blocks import FormStepBlock, FormStepsBlock
```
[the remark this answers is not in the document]
corysutyak: `CoderedAdvancedFormPage` uses the extended `CoderedFormStepBlock` which inherits from their `FormStepBlock`.  All of our File Fields are using extended versions of their file fields.  The only thing that we might want to do related to this concern is to only use wff as a library and extend their `StreamFormMixin` and `Step`/`Steps` classes in the CMS so there is no direct contact between the CMS and wff.

coderedcms/blocks/__init__.py
```diff
@@ -64,3 +69,28 @@
...
+]
+
+
+ADVANCEDFORM_STREAMBLOCKS = ADVANCEDFORM_FORMBLOCKS + CONTENT_STREAMBLOCKS
+
+class CoderedAdvancedFormStepBlock(FormStepBlock):
+    form_fields = StreamBlock(ADVANCEDFORM_STREAMBLOCKS)
+
+
+class CoderedAdvancedFormStepsBlock(FormStepsBlock):
```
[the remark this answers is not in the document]
corysutyak: We would get a circular dependency error if it was in `advanced_form_blocks.py` as we need the Content/HTML block list to build the `CoderedAdvancedFormStepBlock`.  Alternatively, we can define `CoderedAdvancedFormStepBlock` in `advanced_form_blocks.py` but we would have to repeat the Content/HTML block list in that file.

coderedcms/project_template/project_name/settings/base.py
corysutyak: As the code has their own models and hooks, it needs to be defined here.  The model issue will be rectified if we treat wff as a library and extend everything in it.  But we might need to repeat the code in their hooks.

coderedcms/templatetags/coderedcms_tags.py
corysutyak: https://github.com/coderedcorp/coderedcms/pull/171/files#diff-e8b7a6a93ddcb52b42325f9308a07806R400

I didn't have too much time to find the reasoning, but it appears that the above line is what causes the rendering of the fields.  We're extending this model with a proxy model though, so it is fixable.

coderedcms/models/page_models.py
corysutyak: Yeah, but it then gets overwritten again by the `StreamFormMixin`...I'll have to think of a cleaner solution.

coderedcms/blocks/advanced_form_blocks.py
```diff
@@ -0,0 +1,88 @@
...
+
+class CoderedAdvancedFormDropdownFieldBlock(form_blocks.DropdownFieldBlock, FormBlockMixin):
+    pass
+
+
+class CoderedAdvancedFormCheckboxesFieldBlock(form_blocks.CheckboxesFieldBlock, FormBlockMixin):
+    pass
+
+
+class CoderedAdvancedFormDateFieldBlock(form_blocks.DateFieldBlock, FormBlockMixin):
```
vsalvino: Not sure if the problem is here, but the Date/Time/Datetime blocks are not rendering with the correct html `type=""` attribute. They are all showing `type="text"` and seem to have some additional data attributes added to them. Ideally they would use the HTML5 input types to be consistent with how our current forms work.

coderedcms/wagtail_flexible_forms/blocks.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def get_field_class(self, struct_value):
+        text_format = struct_value['format']
+        if text_format == 'url':
+            return forms.URLField
+        if text_format == 'email':
+            return forms.EmailField
+        if text_format == 'phone':
+            return PhoneNumberField
+        return super().get_field_class(struct_value)
```
vsalvino: Interestingly enough, these fields are not rendering with the correct HTML5 input type (email, tel). They are all rendering as `type="text"`. Not sure if that's a problem here, in the bootstrap4 package, or elsewhere.

vsalvino:
Found a bug that occurs when trying to use with a confirmation email. The confirmation email seems to be triggered on step 2, and cuases a 500 error, if a field referenced in the confirmation email is not present in step 2 (e.g. the field is from step 1). Regardless, confirmation emails should not be triggered until the end when the `FormSubmission` is created.

```
Traceback:

File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/local/lib/python3.7/site-packages/wagtail/core/views.py" in serve
  24.     return page.serve(request, *args, **kwargs)

File "/usr/local/lib/python3.7/site-packages/coderedcms/models/page_models.py" in serve
  1612.                     processed_data=submission.get_data()

File "/usr/local/lib/python3.7/site-packages/coderedcms/models/page_models.py" in process_form_submission
  1230.                 message.send()

File "/usr/local/lib/python3.7/site-packages/django/core/mail/message.py" in send
  291.         return self.get_connection(fail_silently).send_messages([self])

File "/usr/local/lib/python3.7/site-packages/django_sendmail_backend/backends.py" in send_messages
  23.             if self._send(message):

File "/usr/local/lib/python3.7/site-packages/django_sendmail_backend/backends.py" in _send
  43.                 raise Exception('send_messages failed: %s' % error)

Exception Type: Exception at /stream-form/
Exception Value: send_messages failed: b'sendmail: No recipients specified although -t option used\n'
```

vsalvino:
Streamform field data is not exactly normalized. I know that we are normalizing file uploads to some extent, but ideally everything should be normalized to remain compatible with how the existing `FormSubmission` works (everything is just plain text, files are just a path) For example, if you try to use the contents of an email field in a ConfirmationEmail, the rendered output is (notice the `To` header):

```
Content-Type: text/html; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Form
From: webmaster@localhost
To: &lt;a href=&quot;mailto:test@example.com&quot;
 target=&quot;_blank&quot;&gt;test@example.com&lt;/a&gt;
Date: Tue, 14 May 2019 22:30:56 -0000
Message-ID: <155787305649.15924.6981146687638059061@CR-WS-5NQYFC2>

Hello My Name!
-------------------------------------------------------------------------------
```

PR #177, opened 2019-05-05
Update issue templates

corysutyak:


vsalvino:
I like these new templates. I would just like to see a 3rd template for Questions/Inquiries/Community feedback, or something along those lines. We could also label those with a special label to easily identify them and ignore them from project planning.

PR #192, opened 2019-05-16
Provide sass starter template

vsalvino:
(Not necessarily ready for merge yet)

This change does a couple things:
* Distribute bootstrap and jquery with coderedcms.
* Allow optionally specifying a starter template via `coderedcms start -t [basic|sass]`
* Moved `project_template/` to `project_template/basic/`.
* Provide new Sass starter template at `project_template/sass/`.
* Provide a default base.html in both basic and sass start templates, since nearly every project ends up needing one.

Thoughts and opinions? It does appear to bloat the project more, but it does seem more proper to distribute the required static assets with the project. When using sass, it removes 2 additional render-blocking resources by bundling all three CSS files into one.

Some alternative ideas:
* Provide bootstrap in the client project / starter template, and leave coderedcms without bootstrap or just relying on the CDN.
    * This would then align with the idea of making starter templates for non-bootstrap styles
* OR: decouple bootstrap from coderedcms into a coderedcms-bootstrap theme, and follow WordPress in pre-installing the the default theme with the starter template. This would require refactoring coderedcms to be "theme-less" on its own (i.e. the html templates would be ultra bare-bones)
* OR: continue with our current philosophy of shipping with bootstrap out of the box, as a core part of coderedcms. This is the simplest option of course. (in which case this PR would likely be merged)


vsalvino:
Ok, this is ready for review. I'm making the executive decision to include the bootstrap files in coderedcms. We can reassess in the future if we want to make downloadable templates/themes. But since bootstrap is deeply integrated in our project, it makes sense to distribute those files.

I made a few recent commits which provide docs and unit tests. Kevin's new code coverage pipeline got me with 0.08% code coverage decrease.

Please review and let me know what you think.

PR #193, opened 2019-05-21
Add basic codecov files

WMeldon:
```codecov``` is installed by travis because it's never used outside the CI environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,12 @@ python:
   - "3.7"
 install:
   - pip install -e './[dev]'
+  - pip install codecov
 before_script:
   - coderedcms start testproject
 # command to run tests
 script:
-  - python testproject/manage.py test coderedcms --settings=coderedcms.tests.settings
-  - flake8 coderedcms testproject
+  - coverage run testproject/manage.py test coderedcms --settings=coderedcms.tests.settings
+  - flake8 coderedcms testproject --exit-zero
+after_success:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: setup.py [hunk 1]
```diff
@@ -54,6 +54,7 @@
     ],
     extras_require={
         'dev': [
+            'coverage',
```
vsalvino: Please make separate extras_require definition for the ci. For example, the ci does not need libsass or sphinx, and the developer does not need coverage, etc.

.travis.yml
```diff
@@ -7,9 +7,12 @@ python:
...
 before_script:
   - coderedcms start testproject
 # command to run tests
 script:
-  - python testproject/manage.py test coderedcms --settings=coderedcms.tests.settings
-  - flake8 coderedcms testproject
+  - coverage run testproject/manage.py test coderedcms --settings=coderedcms.tests.settings
+  - flake8 coderedcms testproject --exit-zero
+after_success:
+  - codecov
```
vsalvino: Is after success mean if there were not test/flake errors? Or does it always run assuming the script didn't bomb out?

WMeldon: It won't run if the tests fail.  flake errors don't cause issues.

vsalvino: Looks good. Reminder to remove the `--exit-zero` flag in the future once we get the code cleaned and flaked.

PR #197, opened 2019-05-23
adding language to html tag and i18n

hayleyhartman:
in settings.py
-changed i18n to True so that international users can access admin in preferred language

in base.py
-changed LANGUAGE_CODE to LANGUAGES list with tuples of possible project languages (leaving 'en' as default)
- to change the default page language, remove the 'en' tuple and add in your preferred language

in base.html
-used i18n to get_current_language from user's browser, and if it is one of the languages in LANGUAGES in base.py, the browser will use that language in the html tags
-if user's current language is not available to that project, the language set in base.py will stay set in the html tags

[django translation doc](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/)

### DIFF
--- a/coderedcms/project_template/project_name/settings/base.py
+++ b/coderedcms/project_template/project_name/settings/base.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/{{ docs_version }}/ref/settings/
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
+from django.utils.translation import gettext_lazy as _
 
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 BASE_DIR = os.path.dirname(PROJECT_DIR)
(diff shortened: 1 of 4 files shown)


vsalvino:
Few points of feedback:
* Test failed. This means there is a bug in the code. Please review the test output and fix so that the tests pass.
* Please create new documentation on how to se the language. Docs are in the `docs/` directory. I'm thinking we create a Recipes section for "how-to" style documentation - this would be a great candidate for our first recipe.

coderedcms/project_template/project_name/settings/base.py
```diff
@@ -138,15 +138,17 @@
...
 
-
 # Internationalization
 # https://docs.djangoproject.com/en/{{ docs_version }}/topics/i18n/
 
-LANGUAGE_CODE = 'en-us'
+# To add or change language of the project, modify the list below.
+LANGUAGES = [
+    ('en', _('English'))
+]
```
vsalvino: After reading the Django docs, I'm thinking we should leave `LANGUAGE_CODE` as-is. It seems like that is the fallback when Django can't figure out what language to use.

PR #199, opened 2019-05-24
drop py34

auvipy:


vsalvino:
Thanks for the PR @auvipy. We have still been experimenting with continuous integration and ended up switching to Azure Pipelines.

PR #206, opened 2019-06-03
0.15.0 release

vsalvino:
Updating release notes... think I got everything.


corysutyak:
Do we want this merged even though the build is failing?

vsalvino:
Lol the build is failing because of a bug in azure pipelines 😠, their current version of python 3.6 is not affected though so that build is more indicative of actual results.

I also just pushed one more commit, was doing some testing and realized there was an endless migration problem due to how we are running the project_template migration files through Django template renderer, and there was a string in there that contained `{{ }}`, hence it was getting stripped and triggering another migration.

PR #207, opened 2019-06-05
Feature/196 title iframe blocks

### LINKED ISSUE #196: Add title field to iFrame blocks
**Describe the bug**
Lack of title for iFrame elements consistently gets reported as an ADA compliance issue when running CMS sites through ADA checker

**Expected behavior**
Add title field to iFrame block to assist disabled folks to know what a video is going to show them
<iframe title="Campus Tour" src="campustourvideo.html"></iframe>

corysutyak:
Fixes #196 

From embedly, we receive the html of the embed and the title of the embed's web page.  Before rendering the html, we use beautifulsoup to inject the title tag into the iframe.

Also removes two unused templatetags and creates almost amp compliant iframes.

### DIFF
--- a/coderedcms/templatetags/coderedcms_tags.py
+++ b/coderedcms/templatetags/coderedcms_tags.py
@@ -2,7 +2,7 @@ import string
 import random
 from html import unescape
 
-
+from bs4 import BeautifulSoup
 from datetime import datetime
 from django import template
 from django.conf import settings
@@ -21,24 +21,6 @@ from coderedcms.models import Footer, Navbar
 from coderedcms.settings import cr_settings, get_bootstrap_setting
 
 register = template.Library()
-
-@register.filter
-def get_embed_video_provider(url):
-    if 'youtu.be' in url or 'youtube.com' in url:
-        return 'youtube'
-    if 'vimeo.com' in url:
-        return 'vimeo'
-    return ''
-
-@register.filter
-def get_embed_video_code(url):
-    if get_embed_video_provider(url) == 'youtube':
-        v = url.split('v=', 1)[1]
-        return v.split('&', 1)[0]
-    if get_embed_video_provider(url) == 'vimeo':
-        v = url.split('.com/', 1)[1]
-        return v.split('?', 1)[0]
-    return ''
 
 @register.filter
 def is_advanced_setting(obj):
@@ -142,7 +124,7 @@ def structured_data_datetime(dt):
     return datetime.strftime(dt, "%Y-%m-%d")
 
 @register.filter
-def richtext_amp(value):
+def richtext_amp_formatting(value):
 
     if isinstance(value, RichText):
         value = richtext(value.source)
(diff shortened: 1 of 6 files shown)


coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -151,3 +133,21 @@ def richtext_amp(value):
 
     value = utils.convert_to_amp(value)
     return mark_safe(value)
+
+@register.filter
+def amp_formatting(value):
```
vsalvino: Could this function be combined with `richtext_amp_formatting`? They seem to do the same thing, and it already checks for RichText instance?

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -151,3 +133,21 @@ def richtext_amp(value):
...
+    return mark_safe(utils.convert_to_amp(value))
+
+@register.simple_tag
+def render_iframe_from_embed(embed):
+    soup = BeautifulSoup(embed.html, "html5lib")
+    try:
+        iframe_tags = soup.find('iframe')
+        iframe_tags['title'] = embed.title
+        return mark_safe(str(soup.body.iframe))
```
vsalvino: Returning `soup.body.iframe` seems like it might miss something, e.g. what if `embed.html` contains a div wrapper around an iframe; then you would lose the wrapper div, right? Couldn't you just return the whole `soup`, or does that add a bunch of unwanted HTML markup?

coderedcms/templatetags/coderedcms_tags.py
corysutyak: I could, but it wouldn't be really clean.  We're not quite checking whether the value is a `RichText` object.  When we need to ampify a rich text block, we need the rendered html from the `RichText` object.  We're not checking if the value in `richtext_amp_formatting` is or is not richtext, it assumes that.  

What we're checking for is whether the value is the internal richtext source (a pseudo template-like state) or a `RichText` object that typically wraps around the source.  We need the source to pass it through the `richtext` function which renders the source.  Then we pass through the fully rendered richtext html to `convert_to_amp`

The better solution would be to make `richtext_amp_formatting` a wrapper around `amp_formatting`, which I'll go ahead and do.

coderedcms/utils.py
```diff
@@ -44,7 +44,7 @@ def convert_to_amp(value):
     Function that converts non-amp compliant html to valid amp html.
     value must be a string
     """
-    soup = BeautifulSoup(value)
+    soup = BeautifulSoup(value, "html5lib")
```
vsalvino: Is there a reason this is using "html5lib" and coderedcms_tags.py is using different "html.parser"?

corysutyak: They should both be using html.parser actually.  html5lib makes it valid html (ie includes body/head tags)

vsalvino:
Alright, looks good to me. Just please the CI gods first. The 3.5 and 3.7 jobs are borked, but the 3.6 job should pass... looks like there are some flake8 errors.

https://dev.azure.com/coderedcorp/coderedcms/_build/results?buildId=89&view=logs&j=1912c076-accd-564d-e4de-1dd031d7cfaf

You can run the flake8 job locally using powershell or pwsh core with `./ci/run_flake8.ps1`. It will only check the diff, not the whole project. We are lacking some docs for the CI stuff still.

corysutyak:
Flake should be good now

PR #210, opened 2019-06-19
Form Spam Protection

### LINKED ISSUE #4: Form Captcha
Ability to enable a caption on forms.

This could potentially be implemented as either a form field, or at the page level. Potentially useful library https://github.com/praekelt/django-recaptcha

corysutyak:
Fixes #4 

This PR adds a boolean to form/stream form pages to toggle spam protection using the suggested spam protection technique provided in the issue.

Additionally, I refactored the serve method on form/stream forms to be more unified.

With the way this spam protection is implemented, if a developer wanted to use their own spam catching mechanism, it should be a fairly straightforward process.


coderedcms/templates/coderedcms/includes/form_honeypot.html
```diff
@@ -0,0 +1,4 @@
+<div style="overflow:hidden;width:0;height:0;" aria-hidden="true">
+    <label for="cr-decoy-comments">Leave this blank if you're a human</label>
```
vsalvino: Should be translatable.

coderedcms/models/page_models.py
```diff
@@ -1327,6 +1332,67 @@ def serve_submissions_list_view(self, request, *args, **kwargs):
...
+
+    def process_spam_request(self, form, request):
+        """
+        Called when spam is found in the request.
+        """
+        messages.error(request, self.get_spam_message())
+        return self.process_form_get(form, request)
+
+    def get_spam_message(self):
+        return "Spam detected, form was not submitted."
```
vsalvino: Should be translatable. I could see this message being taken the wrong way by a human (i.e. we are telling you that you are spam). Maybe update to read: "There was an error while processing your submission. Please try again."

coderedcms/models/page_models.py
```diff
@@ -1327,6 +1332,70 @@ def serve_submissions_list_view(self, request, *args, **kwargs):
...
+            return True
+        return False
+
+    def process_spam_request(self, form, request):
+        """
+        Called when spam is found in the request.
+        """
+        messages.error(request, self.get_spam_message())
+        ip_address = utils.get_ip_from_request(request)
+        logger.info("{0} - Detected spam submission on page: {1}".format(ip_address, self.title))
```
vsalvino: Should probably log the full request for the most helpful info. Plus if there is a false positive the submission would still be captured in the log. You wouldn't even need to format the IP address then because it would be contained in the request headers/info. I'm thinking something similar to the default django error message which spits out the request object (not sure how they do it, but you could probably steal that code).

coderedcms/templates/coderedcms/pages/base.html
```diff
@@ -124,7 +124,15 @@
...
-
+                {% block messages %}
+                    {% if messages %}
+                        <div class="container">
+                        {% for message in messages %}
+                            <div class="alert alert-{{ message.tags|map_to_bootstrap_alert }}" role="alert">{{ message }}</div>
+                        {% endfor %}
+                        </div>
+                    {% endif %}
+                {% endblock %}
```
vsalvino: Nice addition.

coderedcms/templates/coderedcms/pages/form_page.html
```diff
@@ -10,6 +10,13 @@
 <div class="container">
     <form class='{{ page.form_css_class }}' id='{{ page.form_id }}' action="{% pageurl self %}" method="POST" {% if form|is_file_form %}enctype="multipart/form-data"{% endif %}>
         {% csrf_token %}
+
+        {% block captcha %}
+            {% if page.spam_protection %}
+                {% include 'coderedcms/includes/form_honeypot.html' %}
+            {% endif %}
+        {% endblock %}
```
vsalvino: Since we are letting people override this, maybe we should put it at the bottom of the form, where a captcha normally would go. Less of a chance the template would have to be forked then just to change the layout if using something like the google captcha instead.

PR #214, opened 2019-07-09
article page twitter card, image template tag

hayleyhartman:
just wanted to make a pr so you can see what the difference is here. the things in article page work and I think are ready, unless you have any suggestions. The changes in base.html and coderedcms tags are NOT working and I need some python direction for those.

### DIFF
--- a/coderedcms/templatetags/coderedcms_tags.py
+++ b/coderedcms/templatetags/coderedcms_tags.py
@@ -40,6 +40,12 @@ def coderedcms_version():
 def generate_random_id():
     return ''.join(random.choice(string.ascii_letters + string.digits) for n in range(20))
 
+@register.simple_tag
+def twitter_card_image(self):
+    if self.cover_image:
+        return self.cover_image
+    elif settings.coderedcms.LayoutSettings.logo:
+        return settings.coderedcms.LayoutSettings.logo
 
 @register.simple_tag
 def is_menu_item_dropdown(value):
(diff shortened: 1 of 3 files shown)


coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -40,6 +40,12 @@ def coderedcms_version():
 def generate_random_id():
     return ''.join(random.choice(string.ascii_letters + string.digits) for n in range(20))
 
+@register.simple_tag
+def twitter_card_image(self):
```
vsalvino: `self` is a special keyword in Python. In the context of a python file, `self` refers to the actual file. Inside of a `class`, self would refer to the class (it's like `this` in JavaScript or other languages). So in this context, `self` refers to `coderedcms_tags`. Whereas in the template, `self` was referring to the page model.

Try using a different variable name, such as `page` (since you are expecting a page object).

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -40,6 +40,12 @@ def coderedcms_version():
...
+    if self.cover_image:
+        return self.cover_image
+    elif settings.coderedcms.LayoutSettings.logo:
+        return settings.coderedcms.LayoutSettings.logo
```
vsalvino: I don't think `settings` is defined within the current context. Once again, the context of the template, there is a `settings.coderedcms` object holding the settings, however in the python file, that is not defined.

In wagtail, settings are just defined as normal models. So if you need the layout settings you can import them:
```python
from coderedcms.models import LayoutSettings
```
(I think that's where there are, you will have to double check). You can read the wagtail docs about how to use the wagtail settings: https://docs.wagtail.io/en/stable/reference/contrib/settings.html.  The docs will show you the difference between using them in python versus using them in the template.

---
Sort of unreleated, but also of note: The word "settings" gets used a lot and is ambiguous. So within coderedcms, we refer to Django settings (i.e. things defined in settings.py or settings/base.py, etc.) and Wagtail settings (things defined using a wagtail settings model that are then changeable by the user in the wagtail admin)

PR #215, opened 2019-07-10
twitter card adjusted to render images

hayleyhartman:
- added meta tags to base to comply with twitter requirements and called up new template tag, twitter_card_image
- twitter_card_image first tries to use cover image in twitter card, if not uses website logo
- per request, cleaned up the Open Graph image meta tag logic to check for OG image, then cover photo, then logo in the same way as the twitter_card_image.
- article page runs also runs twitter_card_image and adjusts the twitter card to use summary_large_image

### DIFF
--- a/coderedcms/templates/coderedcms/pages/article_page.html
+++ b/coderedcms/templates/coderedcms/pages/article_page.html
@@ -30,7 +30,6 @@
 {% if settings.coderedcms.SeoSettings.twitter_meta %}
     {% block twitter_card %}{% if self.cover_image %}summary_large_image{% else %}{{block.super}}{% endif %}{% endblock %}
     {% block twitter_seo_extra %}
-    {% image self.cover_image fill-2000x1000 as cover_image %}
         <meta name="twitter:title" content="{{ self.title }}">
         {% if self.caption %}
         <meta name="twitter:description" content="{{ self.caption }}">
--- a/coderedcms/templates/coderedcms/pages/article_page.html
+++ b/coderedcms/templates/coderedcms/pages/article_page.html
@@ -30,11 +30,9 @@
 {% if settings.coderedcms.SeoSettings.twitter_meta %}
     {% block twitter_card %}{% if self.cover_image %}summary_large_image{% else %}{{block.super}}{% endif %}{% endblock %}
     {% block twitter_seo_extra %}
-        <meta name="twitter:title" content="{{ self.title }}">
         {% if self.caption %}
         <meta name="twitter:description" content="{{ self.caption }}">
         {% endif %}
-        <meta name="twitter:image" content="{{self.get_site.root_url}}{{ twitter_card_image self }}">
         {% if self.author.twitter %}
             <meta name="twitter:creator" content="{{self.author.twitter}}" />
         {% endif %}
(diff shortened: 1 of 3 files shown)


coderedcms/templates/coderedcms/pages/article_page.html
```diff
@@ -30,6 +30,11 @@
 {% if settings.coderedcms.SeoSettings.twitter_meta %}
     {% block twitter_card %}{% if self.cover_image %}summary_large_image{% else %}{{block.super}}{% endif %}{% endblock %}
     {% block twitter_seo_extra %}
+        <meta name="twitter:title" content="{{ self.title }}">
+        {% if self.caption %}
+        <meta name="twitter:description" content="{{ self.caption }}">
+        {% endif %}
+        <meta name="twitter:image" content="{{self.get_site.root_url}}{{ twitter_card_image self }}">
```
vsalvino: I don't think we need to provide the `twitter:title` or `twitter:image` here, they are already provided in `base.html`

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -40,6 +41,26 @@ def coderedcms_version():
 def generate_random_id():
     return ''.join(random.choice(string.ascii_letters + string.digits) for n in range(20))
 
+@register.simple_tag(takes_context=True)
+def twitter_card_image(context, page):
```
vsalvino: You can scrap this function and use `og_image` function for both twitter card image and opengraph image tags. They should be using the same logic, so I think we just need one function for both.

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -40,6 +41,26 @@ def coderedcms_version():
...
+
+@register.simple_tag(takes_context=True)
+def og_image(context, page):
+    if page.og_image:
+        return page.og_image.get_rendition('original').url
+    elif page.cover_image:
+        return page.cover_image.get_rendition('original').url
+    elif LayoutSettings.for_site(context['request'].site).logo:
+        layout_settings = LayoutSettings.for_site(context['request'].site)
+        return layout_settings.logo.get_rendition('original').url
```
vsalvino: It would be nice if this function returned the full URL (`http://example.com/image.jpg`) rather than just the image URL (`/image.jpg`)

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -40,6 +41,18 @@ def coderedcms_version():
...
+def og_image(context, page):
+    site_url = context['request'].site.root_url
+    if page.og_image:
+        relative_path = page.og_image.get_rendition('original').url
+    elif page.cover_image:
+        relative_path = page.cover_image.get_rendition('original').url
+    elif LayoutSettings.for_site(context['request'].site).logo:
+        layout_settings = LayoutSettings.for_site(context['request'].site)
+        relative_path = layout_settings.logo.get_rendition('original').url
+    return site_url + relative_path
```
vsalvino: `relative_path` is not necessarily defined if none of the if/else statements are ever triggered. In the case where there is no og_image, no cover_image, and no logo.

In that case, you should probably just return an empty string.

Issue #202, opened 2019-05-28
Navbar and Footer snippets to not support multisite


vsalvino:
**Describe the bug**
All navbars and footers are rendered on the page, regardless of the site. We should be specifying the site from the request when querying navbars and footers.

https://github.com/coderedcorp/coderedcms/blob/a3ab4de00e043b22975c06bc5408e9fe2d3a9118/coderedcms/templatetags/coderedcms_tags.py#L81-L87

**To Reproduce**
1. Create multiple sites / root pages
2. Create a different navbar or footer snippet in each site.
3. When browsing each specific site, navbars/footers from both sites are rendered.

**Expected behavior**
Only navbars and footers for a specific site should render when browsing that site.

**Additional context**
See community issue #200

carlosmonari:
Hi everybody,

I wrote a solution for this problem:

Is it possible put my solution in CodeRed CMS ?

I change, and test, my code for NavBar.

The solution is:

in file wagtailsettings_models.py (add after field favicon in class LayoutSettings )

    navbar_name = models.CharField(
        blank=True,
        max_length=50,
        verbose_name=_('Navbar name'),
        help_text=_('It is necessary to associate when you create a Navigation Bar fragments.'),
    )
		
and add panels


# before
    panels = [
        MultiFieldPanel(
            [
                ImageChooserPanel('logo'),
                ImageChooserPanel('favicon'),
            ],
            heading=_('Branding')
        ),
        MultiFieldPanel(
            [
                FieldPanel('navbar_color_scheme'),
                FieldPanel('navbar_class'),
				........
				
# after
    panels = [
        MultiFieldPanel(
            [
                ImageChooserPanel('logo'),
                ImageChooserPanel('favicon'),
            ],
            heading=_('Branding')
        ),
        MultiFieldPanel(
            [
                FieldPanel('navbar_name'),
                FieldPanel('navbar_color_scheme'),
                FieldPanel('navbar_class'),
				........

				
in file snippet_models.py (add after field name in class Navbar )				

# included 
from coderedcms.models.wagtailsettings_models import LayoutSettings

				
    layoutsetting = models.ForeignKey(
        LayoutSettings,
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+',
        verbose_name=_('Layout Page'),
        help_text=_('The Layout page users to determinate a site owner Navigation Bar Manu.'),
    )


and increase panels

# before
    panels = [
        FieldPanel('name'),
[...]

PR #216, opened 2019-07-12
Partial correct [BUG] #202 - Correct a Navbar snippets to support multisite coderedcorp

carlosmonari:
Partial correct [BUG] Navbar and Footer snippets to not support multisite coderedcorp#202

This item correct a Navbar multisite


vsalvino:
Thanks for the PR! We'll review it and leave some feedback.

coderedcms/models/snippet_models.py
```diff
@@ -224,6 +226,15 @@ class Meta:
         max_length=255,
         verbose_name=_('Name'),
     )
+    layoutsetting = models.ForeignKey(
```
vsalvino: Would it make more sense to have a Foreign Key to a `Site` rather than a settings object?

coderedcms/models/wagtailsettings_models.py
```diff
@@ -117,6 +117,12 @@ class Meta:
         related_name='favicon',
         verbose_name=_('Favicon'),
     )
+    navbar_name = models.CharField(
```
vsalvino: I don't think this field is necessary here. The `Navbar` object already has a field for the name.

The navbars are a bit confusing in coderedcms. There is one `<nav>` element (which is what all of these settings pertain to). Within that `<nav>` it can render multiple `Navbar` snippets.

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -67,8 +67,11 @@ def get_pictures(collection_id):
 
 
 @register.simple_tag
-def get_navbars():
-    return Navbar.objects.all()
+def get_navbars(request):
+    curr_url = request.site.hostname
+    site = Site.objects.get(hostname=curr_url)
+    layout = LayoutSettings.objects.get(site_id=site)
+    return Navbar.objects.filter(layoutsetting_id=layout.id)
```
vsalvino: Ditto comment above. If the `Navbar` were to have a foreign key to a `Site`, you could eliminate the LayoutSettings here and filter the navbar directly by the site, e.g.:

```python
Navbar.objects.filter(site=site)
```

rstienstra:
Do we need carlosmonari to respond to the PR: Awaiting Author..?

vsalvino:
@rstienstra Yes the PR needs a little more work. If you want to work on it, I will merge it into a branch which will be easier to fork and have more folks work on it.

rstienstra:
I am looking in to his work and your comments... If I understand it better I can give it a try!

vsalvino:
I merged this into branch: `merge/pr-216`. If you want to work on it, you can fork and start from that branch, and make a pull request back into that branch on coderedcms.

vsalvino:
Closing this as we now have an active branch going for PR #408, which will likely be released in version 0.21.

Will credit @carlosmonari as co-author since he has done a lot of work in this PR which provided the basis. Thanks everyone!

PR #227, opened 2019-08-03
Pipeline improvements

vsalvino:
Use multi-stage pipeline to avoid race condition when comparing code coverage (problem that currently only affects master branch).

Add documentation build to pipeline as well. If there are any warnings such as an incorrect link or a page that is not referenced, it will fail.

Codifying the process for running unit tests, flake8, and building docs locally via uniform powershell scripts. This is a simpler way to run them during local development, and also provides some more helpful console output.


vsalvino:
This PR is ready to go, the new multi-stages are pretty nice. Should fix the failing build status on the master branch.

@kevincummings , you'll be interested to know that I tried using the Azure code coverage APIs, but those somehow don't work within the pipeline (but work outside the pipeline). So downloading the artifacts seems to be most reliable.

Issue #125, opened 2019-03-04
Provide fast lookup of common URLs


vsalvino:
Since the `/favicon.ico` URL does not resolve to an actual route, it goes through the wagtail page lookup route, which in turn does multiple database queries (first to get the 404, second to check for any redirects). Since this URL is *constantly* being accessed by all browsers and bots, we should provide a very fast way of resolving that route. My thought is to create a cached view that redirects to the favicon image in the coderedcms settings.

There are many other URLs that are also very frequently crawled by bots, to the point that it can take down a database if a coderedcms site starts getting hit with bots crawling for wordpress sites, because they will all trigger a database lookup.

I propose that we create a cached view that can quickly resolve a variety of common URLS, such as favicon, various WordPress PHP paths, various other CMS login paths, etc. Basically stuff that bots tend to crawl frequently. We can easily obtain and continue updating a list of these URLs from our hosting platform logs.

Granted, all of this could be done effectively on the web server level, but I think it would be great to provide this in CodeRed CMS by default as a security and performance feature to eliminate one additional server configuration step.

FlipperPA:
This would be great to have as a third-party app. I'd be happy to contribute some ideas.

vsalvino:
Good idea, I hadn't thought of that. Honestly it could probably be called `django-robot-urls` since I don't think there would be anything specific to wagtail in there. Just a view that returns straight up 404s based on a list or regex.

FlipperPA:
Good point, I was thinking specifically of Wagtail because of the catch-all URL rule.

vsalvino:
This issue will be somewhat resolved by pull request #133 which will now cache 404 pages to prevent multiple repeated database lookups. However I still think think adding a package as described above would be a good idea.

vsalvino:
Re-labeling this as maintenance since we have addressed a couple of the major cases causing performance degradation as of 0.16 release.

PR #230, opened 2019-08-03
Adding favicon url path and view

vsalvino:
Most requested URL everywhere is /favicon.ico, which 404s currently. This adds a redirect to the selected favicon in the settings.

Also removing the `cache_page` decorator as it is no longer needed since we are defaulting to the wagtail-cache middleware.

Partially implements #125 .

~~Still need to add tests and docs~~


vsalvino:
@corysutyak ready for review.

PR #234, opened 2019-08-05
Test 404 page and built-in URLs

### LINKED ISSUE #218: Unit tests should test 404s and custom URLs
Since we provide a custom 404 template (when DEBUG=False), we should have a unit tests that tests a non-existant URL with DEBUG=False, and ensures the status code is actually 404. This came up because there was a bug in a templates that gets rendered in the 404, which only occurs when there is no page object. Obscure bug but a test would have caught it.

Additionally, we have a few hard-coded URLs in `coderedcms/urls.py`, and `coderedcms/search_urls.py` (both of which are wired up by default in the project_template). Those URLs should be getting tested as well.

kevincummings:
Fixes #218


coderedcms/tests/test_urls.py
```diff
@@ -1,10 +1,102 @@
...
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Filename'], "{0}.ics".format(event_page.slug))
+        self.assertEqual(
+            response['Content-Disposition'],
+            'attachment; filename={0}.ics'.format(event_page.slug)
+        )
+
+    def test_generate_calendar(self):
+        page = WebPage.objects.create(path='/page/', depth=1, title='Page', slug='page')
```
vsalvino: I believe the generate calendar function is intended to be used with an EventPage?

coderedcms/tests/test_urls.py
```diff
@@ -1,10 +1,102 @@
...
+
+        response = self.client.post("/ical/generate/calendar/", {'page_id': page.pk}, follow=True)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Filename'], 'calendar.ics')
+        self.assertEqual(response['Content-Disposition'], 'attachment; filename=calendar.ics')
+
+    def test_ajax_calendar(self):
+        page = EventIndexPage.objects.create(path='/page/', depth=1, title='Page', slug='page')
```
vsalvino: Would it be a better test to create some child event pages as well, so the calendar has something to load?

coderedcms/tests/test_urls.py
```diff
@@ -1,10 +1,153 @@
...
+    def setUp(self):
+        self.client = Client()
+        self.root_page = Page.get_root_nodes()[0]
+
+    def test_generate_single_event(self):
+        event_page = EventPage(
+            path='/single-event/',
+            depth=1,
+            title='Single Event',
+            slug='single-event'
```
vsalvino: Shouldn't the event page have a date or a recurrence associated with it, to properly test the calendar generation?

coderedcms/tests/test_urls.py
```diff
@@ -1,10 +1,153 @@
...
+            {'s': 'Test Search Query'},
+            follow=True
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotEqual(response.context['results'], None)
+
+
+@pytest.mark.django_db
+class TestEventURLs(unittest.TestCase):
```
vsalvino: All of these event tests should probably be moved to `coderedcms.tests.testapp.models` with the Event models.

Issue #194, opened 2019-05-21
[BUG] Form fields allow blank "field type" in wagtail admin


vsalvino:
**Describe the bug**
In the wagtail admin when creating a `CoderedFormPage`, the field type of form fields is allowed to be blank (----------) which passes validation and saves the page properly.

However on the front-end, the page throws a 500 error because no field type is selected.

**To Reproduce**
1. Create a Form Page
2. Add a field
3. Leave the field type as blank (-----------)
4. Publish the page, and view it on the front-end.

**Expected behavior**
Either there should not be a blank option (defaults to single-line text), or the admin should throw an error when trying to save a form page with a blank field type.

**Additional Context**
The order of the form fields / body also seems to be reversed in the wagtail admin (it should be body before form fields, similar to how it renders on the front-end by default). This must be a regression from the refactor to support stream forms.

vsalvino:
Will be released in 0.17

PR #243, opened 2019-08-12
Form field type now defaults to single line text, making 500 error im…

kevincummings:
#194


coderedcms/project_template/basic/website/migrations/0001_initial.py
```diff
@@ -5,7 +5,7 @@
  """
 {% endcomment %}
 {% verbatim %}
-# Generated by Django 2.2.4 on 2019-08-03 16:14
+# Generated by Django 2.2.4 on 2019-08-12 14:12
```
vsalvino: Also need to update the sass template at project_template/sass/

vsalvino: I meant that the whole `0001_initial.py` migration needs to also be copied to sass template. basic and sass should have the same migration. Not just the dates updated (I just commented on that line to make a comment).

coderedcms/project_template/sass/website/migrations/0001_initial.py
```diff
@@ -5,7 +5,7 @@
  """
 {% endcomment %}
 {% verbatim %}
-# Generated by Django 2.2.4 on 2019-08-03 16:14
+# Generated by Django 2.2.4 on 2019-08-12 14:12
```
vsalvino: This file should be identical to `project_template/basic/website/migrations/0001_initial.py`. The same migrations from basic must also be copied to sass.

vsalvino:
@kevincummings This looks good, please just add documentation. Create a new file for 0.17.0 release, note this as a bug fix, and also under upgrade considerations that they will need to makemigrations. You can see an example in 0.15/0.16.0 release notes.

PR #249, opened 2019-08-14
Unit test AMP pages

### LINKED ISSUE #244: Unit tests for AMP pages
With Settings > SEO > AMP turned on, an AMP version of article pages is available by adding the `?amp` query to the URL. We should have a unit tests on article pages to test for this.

This test would fail if you check out tag v0.16.0 as there is a bug in 0.16.0

kevincummings:
Fixes #244

### DIFF
--- a/coderedcms/models/tests/test_page_models.py
+++ b/coderedcms/models/tests/test_page_models.py
@@ -25,6 +25,9 @@ from coderedcms.tests.testapp.models import (
     StreamFormPage,
     WebPage
 )
+from coderedcms.models.wagtailsettings_models import (
+    SeoSettings
+)
 
 
 class BasicPageTestCase():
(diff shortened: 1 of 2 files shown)


coderedcms/models/tests/test_page_models.py
```diff
@@ -161,6 +164,12 @@ class CoderedStreamFormPageTestCase(AbstractPageTestCase, WagtailPageTests):
 class ArticlePageTestCase(ConcreteBasicPageTestCase, WagtailPageTests):
     model = ArticlePage
 
+    def test_amp(self):
+        SeoSettings.amp_pages = True
```
vsalvino: This is not actually saving the setting. See the example in test favicon.

Issue #56, opened 2018-12-15
Make built-in snippets abstract


vsalvino:
If we want to continue using CODERED_FRONTEND_* settings (which I think we do, they will be the foundation of themes), we will have to make everything that uses our streamfield blocks abstract.

The issue is, many coderedcms models use streamfield blocks that pull values from CODERED_FRONTEND_* settings. Which means if you change one of these settings in your project, it triggers a migration in coderedcms. Which is a bad practice and leads to endless migrations.

This fits in with the paradigm we have been following with the page types. Provide beefy abstract models, and provide a very basic concrete model boilerplate in the project template.

One caveat to making the snippets abstract, is that we will no longer be able to provide a snippet chooser, or hard-code templates that pull in navbar/footer snippets. So this will have to be thought through to figure out how to gracefully handle all of this without re-defining everything in the project template (which would be too much boilerplate I think).

Anyhow, this pretty major regarding future stability. Implementing it inevitably cause data loss. But there is no other way to have dynamic blocks and themes. So I think it is necessary.

corysutyak:
I have to read into it more, but we can probably leverage `GenericForeignKey`s for the abstract snippets/choosers. https://docs.djangoproject.com/en/2.1/ref/contrib/contenttypes/#generic-relations

vsalvino:
Also see related issue: #8

auvipy:
what about using django-polymorphic instead of generic relations?

saileshkush95:
I didn't find code snippets for footer can you please help me to figure out.

vsalvino:
@michaelsteigman `LayoutSettings` and `AnalayticsSettings` both support multi-site now. For example, you can use two different Google Analytics codes, two different logos, etc. on the multi-sites.

However, they are not abstract, so you cannot change or extend them. But the good news is these settings are conveniences, and are not "important" in the sense that you can ignore them and the site will still work perfectly fine.

For multi-site situation, I'd highly recommend implementing your own custom navbar and footer, and then implement your own custom settings model. This way, none of those concrete models should get in your way. And you'll probably need a lot of custom business logic anyhow. Then extend our `base.html` and `web_page_notitle.html` templates to add your custom tracking, navbar, etc.

michaelsteigman:
[...]

PR #251, opened 2019-08-16
Support multi-site navbar

vsalvino:
Merge branch of original pull request at #216 .

Will fix issue #202 .

NEEDS WORK before merging.

CC: @rstienstra @carlosmonari


hazho:
hey awesome team;
I'd like to contribute here

rstienstra:
> hey awesome team;
> I'd like to contribute here

Nice @hazho  I hope @vsalvino is seeing this also... the Support multi-site navbar has been a open issue way to long!

vsalvino:
One of the other problems with refactoring the navbar, is that the navbar (and footer) snippets are still concrete within coderedcms (#56 ). This causes a ton of problem with migrations, etc. when customizing the streamfield. But as a band-aid fix we could probably add a navbar chooser streamfield in the **Settings > Layout** area to select which navbars are shown. Since the settings are per site this would work.

So the design would be similar to this PR, although a bit different:
* Leave navbar model as-is.
* Add streamfield (or ManyToMany) LayoutSettings containing:
  * Multi-select ability of all navbars.
* Add a migration which adds all Navbars to this new field in LayoutSettings, to ensure seamless upgrade for existing sites.

Ortentxi:
is there still work on the multi-site navbar? or any solution/workaround?
seems to me a multi-site is not really useful without different navbars. no multi-site footers are acceptable somehow.

vsalvino:
Closing this as it will be implemented via #408 . Thanks to everyone who contributed with ideas and proof of concept code!

Issue #254, opened 2019-08-22
Add option to open link in new tab


vsalvino:
When adding a Button Link block, Image Link block, or a link in the navbar snippet, there should be a checkbox option to open that link in a new tab (adds the `target="_blank"` to the anchor)

Additionally, on our custom rich text block (or maybe under the seo/general settings if it has to be global) add the checkbox option as well, which would say "open all external links in a new tab", and add a wagtail hook to handle it if that option is checked.

PR #255, opened 2019-08-27
Issue 254

hayleyhartman:
When adding a Button Link block, Image Link block, or a link in the navbar snippet, I added the setting to have links open in new tabs and added the target="_blank" or target="_self" attribute to the html.

For the rich text block, I have been trying to adjust links. Right now, the setting is displaying on the Settings/Layout page.  A couple issues:

1. I'm having trouble checking the layout settings in the webhook defining the new tab, which is currently opening a new tab no matter what. Need a way to check before that webhook just fires.
2. I've also noticed that when I uncheck the new tab box and save, the settings page refreshes, and the box is checked again. Is there another place the setting needs to be loaded so that the settings page itself loads with correct settings selected?

Issue #254


vsalvino:
After discussion, let's proceed as so:

* One global setting, in Layout Settings, for opening links in new tabs.
* Remove settings on individual blocks.
* Remove beautiful soup filtering of rich text.
* Handle the addition of target=_blank in javascript, turned on based on the Layout Settings (similar to the track all button clicks google analytics setting).

vsalvino:
Split the work from this PR into two separate PRs: #278 and #279

PR #260, opened 2019-10-01
Fix code coverage comparison

vsalvino:
Build keeps failing when trying to download previous code coverage from master branch.

Updating YAML keywords which seem to have magically changed in Azure without any notice or version change. Also setting code coverage comparison to pass with warnings if any errors occur.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -104,16 +104,16 @@ stages:
     - task: DownloadPipelineArtifact@2
       displayName: 'Download code coverage from current build'
       inputs:
-        source: 'current'
-        path: '$(Agent.WorkFolder)/current-artifacts'
+        buildType: 'current'
+        targetPath: '$(Agent.WorkFolder)/current-artifacts'
         project: '$(System.TeamProjectId)'
         pipeline: '$(System.DefinitionId)'
 
     - task: DownloadPipelineArtifact@2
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -96,28 +96,35 @@ stages:
...
   - job: codecov
-    displayName: Code Coverage
+    displayName: 'Code Coverage'
+    continueOnError: true
     pool:
       vmImage: 'ubuntu-latest'
 
     steps:
+    - pwsh: write-output "project $Env:project, pipeline $Env:pipeline"
+      displayName: 'Helpppppp'
```
corysutyak: Are the multiple p's intentional?

vsalvino: I'm just shooting in the dark at this point trying to figure out what's going on... hence the work in progress label.

vsalvino:
My theory is that this is failing because it is not downloading an artifact since every available master build is "fail" status.

At one point all of the master builds got deleted due to a 30 day retention limit, and we had not merged to master in over 30 days at one point. Every build after that technically failed due to missing the previous master build.

Setting the code coverage to pass with warnings if it is not able to compare it to previous builds. Will try running this in master twice to see if that fixes the problem and the second build is able to download from the first build after this change.

vsalvino:
Cancelling in favor of #261

PR #271, opened 2019-11-19
A more elegant blog preview page with BS4

RNCTX:

#### Description of change
This project already uses BeautifulSoup for other functions, why not generate (more flexible) post previews with it? If a closing `</p>` tag exists within the post body's first 400 chars, it takes the text from within the actual first paragraph and returns it as the preview text.  Else, it strips the tags and returns the first 200 chars with an ellipsis as it did previously.

#### Documentation

Docs do not explain this functionality in detail, no change required.

#### Tests

Simple change, no migrations or additional dependencies, apply patch and restart server


vsalvino:
Thanks for the pull request. I like this idea... however am a little hesitant on this change as-is, because if a blog post has any malformed HTML (for example missing a closing or opening `<p>`), I'm thinking this could potentially produce a really long or possibly garbage preview. For example, it is somewhat common for editors to use embed HTML for the purpose of videos, social media, etc. and these can get fouled up.

I would like to see some testing, for example, if a page has a closing p without an opening p, what would happen? As long as we can handle the edge cases then I would be happy to merge this in.

vsalvino:
Closing due to inactivity. Please re-open if it becomes relevant or there is a new update.

PR #284, opened 2020-01-15
fixing custom_css_class and custom_css_id reference typo

goodtiding5:
Correct typos on custom_css_class and custom_css_id reference in the block templates.


goodtiding5:
Found some more issues with the custom attribute names.  Close this one and create a new pull request.

Issue #287, opened 2020-01-19
Use dynamic image model


pimarc:
#### Describe the bug
Wagtail allows to change the image model. I did that because I wanted media images to be stored in a data based directory structure. Everything is ok in Wagtail but not in Codered. In fact some files are linked directly to the standard image model of wagtail (ex logo).

#### Steps to reproduce
1. Create a new codered project
2. Change the wagtail model as described in the wagtail documentation in order to write to another directory. I attach my application models.py as an example of a custom model. The following line must also be added to base.py "WAGTAILIMAGES_IMAGE_MODEL = 'website.CustomImage'" and then makemigrations and migrate.
3. Do migrations and go to admin
4. Try to set a logo in the layout
5. This produces an error

#### Expected behavior
I corrected the problem myself but I expect it to be corrected in a future release. The correction is quiet simple:
1. add "from wagtail.images import get_image_model_string" to the wagtailsettings_models.py, page_models.py and snippet_models.py files.
2. replace all occurences of 'wagtailimages.Image' with get_image_model_string()
3. do migrations
[models.py.txt](https://github.com/coderedcorp/coderedcms/files/4083618/models.py.txt)

Eveything is working again then

vsalvino:
Thank you for this info! Could you submit a pull request with the fix? Would me much appreciated.

pimarc:
Done

Regards

Pierre

vsalvino:
Closed by PRs #288 and #336

PR #288, opened 2020-01-20
issue when using alternate image model in wagtail

pimarc:
#### Description of change
Correction for the bug #287. Wagtail allows to change the image model. When doing that everything is ok in Wagtail but not in Codered. In fact some files are linked directly to the standard image model of wagtail (ex logo).

#### Documentation
add "from wagtail.images import get_image_model_string" to the wagtailsettings_models.py, page_models.py and snippet_models.py files.
replace all occurences of 'wagtailimages.Image' with get_image_model_string()

#### Tests
Create a new codered project
Change the wagtail model as described in the wagtail documentation in order to write to another directory. I attach my application models.py as an example of a custom model. The following line must also be added to base.py "WAGTAILIMAGES_IMAGE_MODEL = 'website.CustomImage'" and then makemigrations and migrate.
Do migrations and go to admin
Try to set a logo in the layout
This produces an error


vsalvino:
Thank you for this @pimarc. Based on the looks of this change, it will probably cause problems due to bug #56. However this functionality in your pull request is the "correct" way that it should work. I will do some testing.

vsalvino:
Thanks again for doing this work @pimarc . I'm going to merge this in as-is and add a bit more on top of it.

Issue #275, opened 2019-12-27
Custom ID Doesn't Seem to work.


bahoo:
#### Describe the bug
Specifying a custom ID in the UI doesn't show up in the resulting markup.

#### Steps to reproduce
1. Go to the editor.
2. Add a Hero Unit.
3. Specify a custom ID.
4. Save or Publish your page.
5. View source on page.

#### Expected behavior
Custom `id=""` attribute should be set, alongside custom CSS classes ( which do work as expected in my experience )

#### Additional context
This affects hero units in my use, and possibly other block types as well. Best I can tell, there appears to be a mixup between looking for the attribute in `custom_id` versus `custom_css_id`; possibly an artifact of a minor refactor on variable names at some point. It's called `custom_id` in the models ( i.e. `base_blocks.py` ) but referred to as `custom_css_id` in the templates.

I should flag here too, I see a mix of looking at `self.custom_css_id` versus `self.settings.custom_css_id`; e.g.:

```{% if self.custom_css_id %}id="{{self.settings.custom_css_id}}"{% endif %}```

Granted, I haven't spent a ton of time in this area of the codebase so it's possible this is user error or misunderstanding on my part, but if you can provide a thumbs up that I'm understanding the problem correctly, I'm happy to get a PR together.

Thanks!

vsalvino:
What you are describing seems entirely plausible as a bug. I will look into it. If you figure out the exact cause, feel free to submit a pull request with a fix!

topiaruss:
Similarly - I notice that reusable-content elements have only class="block-html", and neither Custom CSS Class nor ID is added.

vsalvino:
Fixed via #293 and #319

PR #293, opened 2020-01-31
Corrects attributes set for custom CSS classes or custom IDs.

bahoo:
#### Description of change
Corrects attributes set for custom CSS classes or custom IDs.

As promised per https://github.com/coderedcorp/coderedcms/issues/275 , unifies the custom CSS and ID attributes. Pulls from `self.settings` consistently, and also uses `custom_id` rather than `custom_css_id`.

### DIFF
--- a/coderedcms/templates/coderedcms/blocks/card_block.html
+++ b/coderedcms/templates/coderedcms/blocks/card_block.html
@@ -1,7 +1,7 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 
 <div class="card mb-3 {{self.settings.custom_css_class}}"
-{% if self.settings.custom_css_id %}id="{{self.settings.custom_css_id}}"{% endif %}>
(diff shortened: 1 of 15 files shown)


topiaruss:
Similarly - I notice that reusable-content elements have only class="block-html", and neither Custom CSS Class nor ID is added.

topiaruss:
@vsalvino Many thanks for this great piece of work! Do you have any thoughts of sprinting to, for example, add block rendering test coverage?

vsalvino:
Normally we would sprint for a day or two after various conferences (PyCon, DjangoCon, Wagtail Space). Since those have become virtual, we kind of lost sprint momentum this year.

There are some basic block rendering tests... but there is a LOT of room for improvement on our tests. If you feel up to the task, I would happily review a PR or help you implement more tests.

We have some basic contributing docs for adding new tests: https://docs.coderedcorp.com/cms/stable/contributing/index.html#adding-new-tests

PR #340, opened 2020-08-02
Use correct Site for LocationPage Google API keys

vsalvino:
Location pages currently use the default site for getting GoogleApiSettings. This should instead use the page's site.


vsalvino:
Abandoning this as it references the wrong issue number.

PR #342, opened 2020-08-02
Use correct site when loading settings

vsalvino:
WIP

Location pages currently use the default site for getting GoogleApiSettings. This should instead use the page's site.


vsalvino:
Tests are currently failing because:
* When a page is initially created, it does not belong to a site.
* It is then added to a page tree, which belongs to a site.
* So any activity that happens on page `save()` cannot rely on a site (therefore site settings) being present.

Not sure how to best handle this, as the Location pages specifically need to perform geocoding on save. Perhaps there is a wagtail hook which will help (i.e. we could geocode after added to page tree versus after saving)

PR #361, opened 2020-08-30
modify position of cover_image block of article_page

python-noob-tk:
#### Description of change
I replaced the cover_image block of article_page.html from outside to inside the container block.
This will fix the incorrect image layout.

### DIFF
--- a/coderedcms/templates/coderedcms/pages/article_page.html
+++ b/coderedcms/templates/coderedcms/pages/article_page.html
@@ -41,11 +41,11 @@
 
 {% block content %}
     <article class="codered-article {%if self.cover_image %}has-img{% endif %}">
+        <div class="container">
         {% if self.cover_image %}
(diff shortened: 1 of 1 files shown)


vsalvino:
Thanks for making a contribution. I'm not sure this is actually a bug, but more a matter of opinion (image is inside of a container versus full width). If there is a bug, can you please explain it further?

python-noob-tk:
Thank you for your comment.

If I use the current template, part of the img will overlap the article with a width of 1200px or more.
I cannot see a part of the image.
There is no control over this situation.

In the case of the modified template, img does not overlap even if the width is 1200px or more.
So I suggested a template fix.

image

python-noob-tk:
I withdraw this proposal.
I'm not sure if my suggestion is correct, so I'll study CSS more.

PR #371, opened 2020-10-28
Added WagtailFormBlock to CONTENT_STREAMBLOCKS

ayushin:
This is very much work in progress, do not merge.


vsalvino:
Closing this due to inactivity. If you need to embed a form, I would recommend using the existing "Page Preview" block, and point it to a form page.

PR #374, opened 2020-11-02
Sending mails through other connection

JorneVL:
This PR allows sending mails through other connections than the default in Django settings
The new connection can be created based on a Site Setting or just hardcoded.

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -1252,9 +1252,7 @@ class CoderedFormMixin(models.Model):
                 message_args['to'] = template_to.render(context).split(',')
 
                 # Send email
-                message = EmailMessage(**message_args)
-                message.content_subtype = 'html'
-                message.send()
+                self.__send_mail(message_args)
 
         for fn in hooks.get_hooks('form_page_submit'):
             fn(instance=self, form_submission=form_submission)
@@ -1293,7 +1291,11 @@ class CoderedFormMixin(models.Model):
             message_args['reply_to'] = template_reply_to.render(context).split(',')
 
         # Send email
+        self.__send_mail(message_args)
+
+    def __send_mail(self, message_args):
         message = EmailMessage(**message_args)
+        message.content_subtype = "html"
         message.send()
 
     def render_landing_page(self, request, form_submission=None):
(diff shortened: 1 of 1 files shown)


vsalvino:
It looks like this change moves the email send into a new private method, but does not actually change the functionality at all. Can you explain your reasoning?

JorneVL:
That is correct, I implemented in a local test version as a public method. But I thought that the private method should be better.
The idea is that you override this method on a parent class (which was possible with my public method).
Do I need to change the method to public so it is possible to override?

coderedcms/models/page_models.py
```diff
@@ -1293,7 +1291,11 @@ def send_summary_mail(self, request, form, processed_data):
             message_args['reply_to'] = template_reply_to.render(context).split(',')
 
         # Send email
+        self.__send_mail(request, message_args)
+
+    def __send_mail(self, resuest, message_args):
```
vsalvino, line 1296: Let's make this public, e.g. `send_mail()` would be fine.

coderedcms/models/page_models.py
```diff
@@ -1293,7 +1291,11 @@ def send_summary_mail(self, request, form, processed_data):
             message_args['reply_to'] = template_reply_to.render(context).split(',')
 
         # Send email
+        self.__send_mail(request, message_args)
```
vsalvino, line 1294: This message is not HTML, it is plain text. So that should probably be an argument to `send_mail()`. In the current change it is hard-coded to HTML-only.

coderedcms/models/page_models.py
vsalvino: Also, it doesn't look like the `request` arg is used/needed here.

vsalvino (changes requested):

coderedcms/models/page_models.py
JorneVL: My use case has the next code, so request can not be removed
```python
mail_settings = MailSettings.for_request(request)
connection = get_connection(
    username=mail_settings.username, password=mail_settings.password, fail_silently=False
)
```

coderedcms/models/page_models.py
JorneVL: I have added an optional parameter (default is text, is that correct?)

vsalvino:
Thanks for your contribution, and your patience, @JorneVL !

This will be merged via #443 and will go out in version 0.22.0

PR #385, opened 2020-11-20
Fix submit form with file uploads for anonymous users

huubbouma:
#### Description of change
Explain what you changed, and how or why it works the way it does.
submitting a form as anonymous user doesn't work. (error 500)
The problem is that the anonymous user doesn't have a session  yet


vsalvino:
Great catch... thanks!

PR #386, opened 2020-11-23
Override the wagtail menu item

JorneVL:
This shows the CoderedCMS version where the original Wagtail Version is placed as this is more important.
With this change, we can remove the additional image 'powered by' and simplify the maintenance of upgrading Wagtail


vsalvino:
Hi Jorne,

Thanks for putting in the effort and making a PR. However we plan to keep the "powered by" section at the bottom, which shows the CodeRed name and version number. As for the settings panel, that used to show the coderedcms version, however we have recently switched that back to show the Wagtail version for consistency with Wagtail.

JorneVL:
Ok, thanks for the information. That sounds good to me

PR #389, opened 2020-12-04
Fix search 500 error with invalid query

### LINKED ISSUE #387: str has no attr 'DoesNotExist' when passing random search_model argument as /search/?s=keyword&t=t
#### Passing random search_model argument as /search/?s=keyword&t=t
Hackers are scanning Codered websites by passing random args for parameter t like this:
 /search/?s=keyword&t=t

This dumps error: 'str' object has no attribute 'DoesNotExist'

I believe this is coming from lib/python3.7/site-packages/coderedcms/views.py
Can we replace **except search_model.DoesNotExist:**
by just **except**
or by, **except ContentType.DoesNotExist:**

rcoldiron:
Fixes #387 

#### Description of change
1) Updated the exception in search to be contentype.doesnotexist so that if someone adds search parameters into the search URL, it returns a 200 page with "no results found" instead of Internal Service Error
2) Added the test for it in the test_search function in test_urls.py

#### Documentation
n/a

#### Tests
Test added to the def test_search function to test for "/search/?s=keyword&t=t"
Also tested this manually and it works. The search box was fine. It was a bug when people typed in search parameters in the URL for search.

### DIFF
--- a/coderedcms/tests/test_urls.py
+++ b/coderedcms/tests/test_urls.py
@@ -47,7 +47,14 @@ class TestSiteURLs(unittest.TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertNotEqual(response.context['results'], None)
 
-        response = self.client.get("/search/?s=keyword&t=t")
+        response = self.client.get(reverse(
+            'codered_search'),
+            {
+                's': 'keyword',
+                't': 't',
+            },
+            follow=False
+        )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['results'], None)
 
--- a/coderedcms/tests/test_urls.py
+++ b/coderedcms/tests/test_urls.py
@@ -47,6 +47,10 @@ class TestSiteURLs(unittest.TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertNotEqual(response.context['results'], None)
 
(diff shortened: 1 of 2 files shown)


coderedcms/tests/test_urls.py
```diff
@@ -47,6 +47,10 @@ def test_search(self):
         self.assertEqual(response.status_code, 200)
         self.assertNotEqual(response.context['results'], None)
 
+        response = self.client.get("/search/?s=keyword&t=t")
```
vsalvino, line 50: Nice test. Could you change the hard-coded URL to use the `reverse()` function to generate it instead (as in line 41 above)

vsalvino (approved):

coderedcms/tests/test_urls.py
rcoldiron: Okay, I updated it. Does follow mean that the search finds the search item? I set this to follow=False to mimic the item not existing but I don't know if that's the right way to do that or if I misunderstood what follow meant.

vsalvino: Follow means if the server returns a redirect (301/302 status code) it will follow the redirect. Otherwise it will stop and give you the 301 status code. Generally it is more useful to follow unless you are specifically testing for a 301/302 redirect.

PR #399, opened 2021-01-16
Generate valid html id

### LINKED ISSUE #398: `generate_random_id` template tag produces invalid id
I've been tracking down a strange issue where modals would randomly not work.  The frustrating part was the same modal that failed would work after refreshing the page.

This is what I have come up with:

[generate_random_id](https://github.com/coderedcorp/coderedcms/blob/a0dd8e9a48ec7141ce8beb87d0113a6c281e15f6/coderedcms/templatetags/coderedcms_tags.py#L42) does not produce a valid html id.  This breaks bootstrap modals.

Per https://www.w3.org/TR/html401/types.html#type-name id must start with a letter `[A-Za-z]`.

Simple fix would be to prepend something like 'cr-' to the id

thenewguy:
Fixes #398

### DIFF
--- a/coderedcms/tests/test_templatetags.py
+++ b/coderedcms/tests/test_templatetags.py
@@ -5,7 +5,7 @@ from django.test import TestCase
 
 
 django_engine = engines['django']
-html_id_re = re.compile("^[A-Za-z][A-Za-z0-9_:\.-]*$")
+html_id_re = re.compile("^[A-Za-z][A-Za-z0-9_:\.-]*$")  # noqa: W605
 
 
 class TemplateTagTests(TestCase):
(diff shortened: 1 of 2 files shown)


coderedcms/tests/test_templatetags.py
```diff
@@ -5,7 +5,7 @@
 
 
 django_engine = engines['django']
-html_id_re = re.compile("^[A-Za-z][A-Za-z0-9_:\.-]*$")
+html_id_re = re.compile("^[A-Za-z][A-Za-z0-9_:\.-]*$")  # noqa: W605
```
vsalvino, line 8: This should probably use a raw string (`r""` ) to avoid the escape sequence confusion.

PR #408, opened 2021-03-01
Multisite navs

rcoldiron:
#### Description of change

- I updated the LayoutSettings model to include a navbar chooser and footer chooser
- I updated the get_footer and get_navbar functions to pull in the choice from settings
- I updated the templates to pull in the choices and also what to do if no choice is made
- Related to #202 
- Multiple sites can use the same navbar/footer if chosen in the Layout Settings for each site
- Simple implementation for select site navbar and footer

#### Documentation
Updated docs with the change.

#### Tests
Manual testing after migrating: Create a navbar and footer snippet and select them in the settings chooser. Make several to switch between snippets and also see what happens when no snippet is chosen. Add an additional site to Sites and select a navbar and footer for that site.


coderedcms/models/wagtailsettings_models.py
```diff
@@ -118,6 +120,15 @@ class Meta:
         related_name='favicon',
         verbose_name=_('Favicon'),
     )
+    navbar_chooser = models.ForeignKey(
```
vsalvino, line 123: I think for this to be more useful it should be a ManyToMany field (allow them to select multiple navbars and footers). An it would then render them in order. This is similar to the current functionality where it renders all of the navbars.

To accomplish this you might need to use modelcluster/orderable, which lets you select multiple and rearrange the order of them. This is similar to how we have multiple carousel slides on a carousel snippet.

rcoldiron:
@vsalvino Ready for another full review. Then show me how to write a manual migration please.

coderedcms/migrations/0020_footerorderable_navbarorderable.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 3.1.7 on 2021-03-12 19:53
+
+from django.db import migrations, models
+import django.db.models.deletion
+import modelcluster.fields
+
+
+class Migration(migrations.Migration):
```
vsalvino, line 8: So what you want to do now is manually select the site's current navbar and footers in the chooser, so that older sites will be converted to the new way of doing it.

For this you can add a `RunPython` operation (the operations are run sequentially) after the other operations are done. This function should loop through the current navbars/footers and create NavbarOrderable and FooterOrderable models which link them up to the LayoutSettings.

See example in the Data Migrations section of django docs: https://docs.djangoproject.com/en/3.1/topics/migrations/#data-migrations

rcoldiron: See new changes below (includes three commits, with final commit the working version). I got it to work for pulling in one instance of nav/footer. What I want to know is: What if we have separated nav snippets manually in a template, if we pulled in all nav snippets, wouldn't that override it? How can we pull in only current navs/footers and not future navs/footers if this migration is run again. That's why I didn't do .all() for it and just did the first navbar or footer. Hopefully I am making sense.

I tested it by creating a new db in dev branch, adding navbars and footers the usual way, then switching to my branch and migrating.  @vsalvino

rcoldiron: Also, new tests will need to be written because this fails pytest really bad....

rcoldiron: Maybe this part isn't necessary? 
`layout.site_navbar = []
 layout.save()`
In case it's not the tests that need updated....

rcoldiron:
NOTES TO SELF: Next steps are to add unit tests and documentation.

Issue #409, opened 2021-03-10
Upgrade to Wagtail 2.13


vsalvino:

thenewguy:
Any timeline for this release?  Running into this issue with embeds https://github.com/wagtail/wagtail/issues/1745 which is resolved in 2.12 per https://github.com/wagtail/wagtail/blob/0daae4a2b829445ee97906dd4393398778881fcb/CHANGELOG.txt#L67

FlipperPA:
@vsalvino Any chance we want to skip directly to 2.13 for 0.21?

vsalvino:
A 2.13 branch is already in progress, based on this branch. If it needs a lot of work we may merge 2.12 in first, if it is a simple update then we will likely skip to 2.13.

vsalvino:
Abandoning in favor of #418

vsalvino:
Merged into dev

PR #412, opened 2021-03-15
Wagtail upgrade

rcoldiron:
For issue #409 

#### Description of change
Updated required Wagtail version.
Also added unidecode in settings since it was not automatically found otherwise, used in blocks.py

#### Tests
Ran the pytest unit tests, all 56 passed with 40 deprecation warnings from Django (some of it from packages).
Also ran the testproject and it all still works, and CSS looks fine too. 
Note that wagtail-import-export is no longer maintained but that will be a separate issue.  It still seems to work for now, though. 

**Will need to change version number but you can test it yourself first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
         'Django>=2.2,<3.2',             # should be the same as wagtail
         'geocoder==1.38.*',
         'icalendar==4.0.*',
-        'wagtail==2.11.*',
+        'wagtail==2.12',
         'wagtailfontawesome>=1.2.*',
         'wagtail-cache==1.*',
         'wagtail-import-export>=0.2,<0.3'
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,8 @@ setup(
         'Django>=2.2,<3.2',             # should be the same as wagtail
         'geocoder==1.38.*',
         'icalendar==4.0.*',
-        'wagtail==2.12',
+        'unidecode==1.*',
+        'wagtail==2.12.*',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -49,7 +49,8 @@
         'Django>=2.2,<3.2',             # should be the same as wagtail
         'geocoder==1.38.*',
         'icalendar==4.0.*',
-        'wagtail==2.11.*',
+        'unidecode==1.*',
```
vsalvino, line 52: Is there a reason for this new dependency?

rcoldiron: When I ran pytest without it, it said ModuleNotFound and broke it. unidecode is used in the blocks.py file. Is there another way to bring it in so that it's recognized as a module?

vsalvino: Ok, I'm guessing wagtail may have had it as a dependency and no longer uses it in 2.12. I would say first let's see where/why we are using it... maybe it can be removed. See if wagtail 2.12 release notes mentions it at all.

rcoldiron: Found a note in 2.11 actually: "Remove unidecode and use anyascii in for Unicode to ASCII conversion" You are currently using it in wagtail_flexible_forms > blocks.py and it's being used in get_slug.

vsalvino: Please make the corresponding replacement to anyascii in our code, if possible, to remain closer to wagtail and eliminate the extra dependency.

rcoldiron: done.

vsalvino:
Abandoning in favor of #418

PR #415, opened 2021-04-08
Fields for Additional Tracking Scripts

rcoldiron:

#### Description of change
Added fields in the Analytics Settings for additional tracking scripts and placed them in the base.html. 
Increased sizing of the boxes in the admin using CSS.

### Documentation
Added a how-to guide on adding the tracking scripts.


coderedcms/templates/coderedcms/pages/base.html
```diff
@@ -196,5 +199,9 @@
             {% endblock %}
             {% block struct_seo_extra %}{% endblock %}
         {% endif %}
+
+        {% if settings.coderedcms.AnalyticsSettings.below_body_scripts %}
```
vsalvino, line 203: Should probably wrap this in a block to make it easy to override if needed. E.g.

```
{% block body_tracking_scripts %}
```

coderedcms/models/wagtailsettings_models.py
```diff
@@ -226,6 +226,16 @@ class Meta:
         verbose_name=_('Track button clicks'),
         help_text=_('Track all button clicks using Google Analytics event tracking. Event tracking details can be specified in each button’s advanced settings options.'),  # noqa
     )
+    below_head_scripts = models.TextField(
+        blank=True,
+        null=True,
+        verbose_name=_('Add tracking scripts between the <head> tags.'),
```
vsalvino, line 232: A bit long for name, might work better as help_text. Please also double check that this renders correctly in the wagtail admin (e.g. `<head>` shows up properly and not as an actual HTML tag)

verbose_name= `<head> tracking scripts`

Ditto for body_scripts

vsalvino (changes requested):

coderedcms/templates/coderedcms/pages/base.html
rcoldiron: Done. Didn't know you could make up block names.

coderedcms/models/wagtailsettings_models.py
rcoldiron: done. it looks like this: 
tracking_analytics_dashboard

coderedcms/fields.py
```diff
@@ -11,3 +12,16 @@ def __init__(self, *args, **kwargs):
     def formfield(self, **kwargs):
         kwargs['widget'] = ColorPickerWidget
         return super(ColorField, self).formfield(**kwargs)
+
+
+class ScriptField(models.TextField):
+    def __init__(self, *args, **kwargs):
+        super(ScriptField, self).__init__(*args, **kwargs)
```
vsalvino, line 19: Since you are not adding any custom logic here, you can skip overriding the `__init__` function. It is essentially doing nothing.

coderedcms/fields.py
```diff
@@ -11,3 +12,16 @@ def __init__(self, *args, **kwargs):
...
+
+    def formfield(self, **kwargs):
+        kwargs['widget'] = Textarea(attrs={
+            'rows': 15,
+            'cols': 10,
+            'style': 'font-family:SFMono-Regular,Menlo,Monaco,Consolas,monospace;'
```
vsalvino, line 25: I would prefer to use a CSS class rather than inline style... I think there is already a class named `monospace` (either in wagtail or coderedcms admin css). If not, we can add one.

coderedcms/fields.py
```diff
@@ -11,3 +12,16 @@ def __init__(self, *args, **kwargs):
...
+
+
+class ScriptField(models.TextField):
+    def __init__(self, *args, **kwargs):
+        super(ScriptField, self).__init__(*args, **kwargs)
+
+    def formfield(self, **kwargs):
+        kwargs['widget'] = Textarea(attrs={
+            'rows': 15,
+            'cols': 10,
```
vsalvino, line 24: I don't think cols is necessary... I believe it fills to 100% width within wagtail admin.

vsalvino (changes requested):

coderedcms/fields.py
rcoldiron: done

coderedcms/fields.py
rcoldiron: done

coderedcms/fields.py
rcoldiron: done. There is a class but I didn't to get specific for it to work.

PR #429, opened 2021-08-02
Django autofields

rcoldiron:
Added settings for Default Auto Fields for Django 3.2

### DIFF
--- a/coderedcms/project_template/basic/project_name/settings/base.py
+++ b/coderedcms/project_template/basic/project_name/settings/base.py
@@ -203,3 +203,7 @@ BOOTSTRAP4 = {
 # Tags
 
 TAGGIT_CASE_INSENSITIVE = True
+
+# Sets default for primary key IDs - change to BigAutoField if you prefer a 64-bit integar instead
+# See https://docs.djangoproject.com/en/3.2/ref/models/fields/#autofield
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
--- a/coderedcms/project_template/basic/project_name/settings/base.py
+++ b/coderedcms/project_template/basic/project_name/settings/base.py
@@ -204,6 +204,7 @@ BOOTSTRAP4 = {
 
 TAGGIT_CASE_INSENSITIVE = True
 
-# Sets default for primary key IDs - change to BigAutoField if you prefer a 64-bit integar instead
-# See https://docs.djangoproject.com/en/3.2/ref/models/fields/#autofield
-DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
+
+# Sets default for primary key IDs
+# See https://docs.djangoproject.com/en/{{ docs_version }}/ref/models/fields/#bigautofield
+DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
--- a/coderedcms/apps.py
+++ b/coderedcms/apps.py
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 
+
 class CoderedcmsConfig(AppConfig):
     name = 'coderedcms'
     verbose_name = 'CodeRed CMS'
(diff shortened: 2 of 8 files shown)


coderedcms/project_template/basic/project_name/settings/base.py
```diff
@@ -203,3 +203,7 @@
 # Tags
 
 TAGGIT_CASE_INSENSITIVE = True
+
+# Sets default for primary key IDs - change to BigAutoField if you prefer a 64-bit integar instead
+# See https://docs.djangoproject.com/en/3.2/ref/models/fields/#autofield
```
vsalvino, line 208: Since these python files are technically templates, the django version in the URL (3.2) should be replaced with the template variable (see other urls in this file)

coderedcms/project_template/basic/project_name/settings/base.py
```diff
@@ -203,3 +203,7 @@
...
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
```
vsalvino, line 209: It looks like as of Django 3.2, BigAutoField is the default, therefore it would make sense that since this template is for brand new projects, we should follow their default.

https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-DEFAULT_AUTO_FIELD

vsalvino, line 209: It looks like new projects generated with Django 3.2 are using `BigAutoField`, so it would make sense for us to follow that, given that these templates are only used for brand new coderedcms projects.

https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

vsalvino:
From reading through the Django docs, I think if someone changes this setting in their project, it would consequently create a migration in coderedcms for any of our concrete models, which could potentially break things.

Therefore I think we should also set this our AppConfig (`app.py`) in coderedcms (and any other of our packages which have concrete database models/migrations) to avoid accidental breakage.

https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

vsalvino:
You may want to add in the release notes too, at the top of upgrade considerations, to add that `DEFAULT_AUTO_FIELD` to your existing project's settings.

coderedcms/project_template/basic/project_name/settings/base.py
rcoldiron: done

coderedcms/project_template/basic/project_name/settings/base.py
rcoldiron: updated in next commit

vsalvino (approved):

rosslaird:
The [release notes](https://docs.coderedcorp.com/cms/releases/v0.21.0.html) indicate that "If you had previously manually specified a different DEFAULT_AUTO_FIELD Django setting, you may need to create a migration." I had *not* manually specified a different DEFAULT_AUTO_FIELD, but I did receive the following warning while running migrations:

    WARNINGS:
    website.EventOccurrence: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
    HINT: Configure the DEFAULT_AUTO_FIELD setting or the WebsiteConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.

Adding `DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'` to my settings (in base.py), as suggested by Vince above, removed the warning. So, if others are experiencing this warning while upgrading, perhaps a note might be added to release notes to indicate the fix.

vsalvino:
Looks like you made the correct change in your project... we should probably update the release notes to indicate that developers should add that setting to their `base.py` to avoid the new Django warning.

Issue #420, opened 2021-06-15
Do not render /contact if it is requested directly by spammers


adminBTI:
Despite the anti-spam honeypot, I keep getting spam emails bothering me to renew my domain name or buy their "contact us" form spam services.

Legitimate users would access my website first and then go to the /contact page to submit the form. How can I display an error page instead of the /contact form for those spammy clients who are hitting my /contact link directly?

Can I use @csrf_protect decoration on the rendering and receiving functions of the /contact form? If so, which functions?

Is it possible to use a different honeypot field for each client?

thenewguy:
How would you detect these spammers?  It would not be unexpected for someone to come to your "Contact Us" page directly from Google.

Perhaps you should look at adding a captcha message to your form if you want to make it more difficult for spammers?  This one is very effective: https://www.google.com/recaptcha/about/

adminBTI:
I exclude /contact page from search engine crawling! Also, in my /contact page, I only display the actual form, no other useful content.

Google's products are meant for its own employees, not for others. If you don't agree with me, then you are still young. I don't know about you, but I often cannot solve these Google Captchas! Is a tiny corner of traffic light box, which is not the actual light, still considered a traffic light? Is the box with just the tip of bicycle handle bar not to be counted? I don't know.

A lot of my website visitors are not from California. So, they would not agree to the jurisdiction of California (a requirement for using Google Captcha. Read their T&C and Privacy policies). A lot of websites are infected with dependent links to Google's fonts, scripts, captchas etc..

This is what I use in my Nginx server: Inside server{..} but outside all location directives:

```
# Assuming the following defaults are not changed:
# CSRF_COOKIE_NAME = 'csrftoken'
# CSRF_HEADER_NAME = 'HTTP_X_CSRFTOKEN'

if ("$uri$cookie_csrftoken$http_x_csrftoken" = "/contact/") { return 404; }
```

thenewguy:
It sounds like you've got it figured out. Best of luck 👍

onaralili:
> I exclude /contact page from search engine crawling! Also, in my /contact page, I only display the actual form, no other useful content.
> 
> Google's products are meant for its own employees, not for others. If you don't agree with me, then you are still young. I don't know about you, but I often cannot solve these Google Captchas! Is a tiny corner of traffic light box, which is not the actual light, still considered a traffic light? Is the box with just the tip of bicycle handle bar not to be counted? I don't know.
> 
[...]

PR #451, opened 2021-10-02
Simple captcha

pppls:

#### Description of change
I added a simple captcha, based on comments in issue #420. @vsalvino mentioned he found a good open-source captcha package to integrate (https://django-simple-captcha.readthedocs.io/en/latest/index.html) and I found wagtail variants in the form of 
https://github.com/acarasimon96/wagtail-django-simple-captcha
and
https://github.com/springload/wagtail-django-recaptcha
The latter has implemented the bulk of the work, the former just makes it so that it works with simple-captcha rather than recaptcha.
I saw no easy way to import those existing package to integrate with coderedcms and rather opted to utilize their code directly (in the case of the CodeRedFormBuilder we could subclass it from [WagtailCaptchaFormBuilder](https://github.com/acarasimon96/wagtail-django-simple-captcha/blob/376bf74a322eb11343709154feae50618184c386/wagtailcaptcha/forms.py#L12) but I'm not sure if this is as trivial for the code in the CoderedFormPage class in the page_models file due to the fact that process_form_submission has been redefined somewhere earlier and because it does not subclass AbstractForm).

I have clearly marked which functions I am talking about **but I would like input on how you prefer to reference their source code**.

The functionality looks as follows:
image
and can be toggled via the spam_protection checkbox under settings.

#### Documentation
I added a small part that says the captcha will be created when toggling the spam_protection checkbox.

#### Tests
Due to not importing the packages themselves directly, we would need to copy the tests as well, but the structure of their [testapp](https://github.com/acarasimon96/wagtail-django-simple-captcha/tree/master/tests/testapp) looks a bit different from what is being done in CodeRedCMS so I'm not sure to what extent I should copy it.


vsalvino:
Thanks for doing this work. Since it's a bigger change it might take a while to be fully reviewed, but it's on our radar. Thanks!

ssharma94:
Hi @vsalvino  I am looking to integrarte this. I can do my version of the coderedcms but it would be awesome that it's a part of the comderedcms package. Could we progress with this?

Issue #457, opened 2021-12-24
CODERED_BANNER does not appear in the CMS


James-Osgood:
#### Describe the bug
The [`CODERED_BANNER` setting](https://docs.coderedcorp.com/cms/reference/django_settings.html#codered-banner) is not applied to the Wagtail CMS.

#### Steps to reproduce
1. Add `CODERED_BANNER = "Test"` to `settings/base.py`
2. Start the server.
3. Sign into the CMS (default is `admin/`).

#### Expected behavior
A light yellow banner at the top of the screen with the text "Test".

#### Additional context
It looks like the CMS banner was removed in #226 with the deletion of the `base.forked.html` template. If this was intentional, then the docs should be updated.

Otherwise, it looks like it would be difficult to add the banner in an appropriate place to the CMS without overriding the base Wagtail admin template again or adding some JavaScript.

vsalvino:
I don't think this removal was intentional. If you're able to make a pull request with the fix that would be appreciated :)

James-Osgood:
Sure. I'll put a PR together to add the banner back.

vsalvino:
Closed via #459

PR #226, opened 2019-08-02
Wagtail 2.6 upgrade

vsalvino:
Spot checked the wagtail admin, not too many changes needed. Fewer than I expected given all of the css and html changes they made to improve accessibility.

PR #459, opened 2022-01-06
Fix `CODERED_BANNER` setting not appearing in CMS

James-Osgood:
#### Description of change
Fix for #457:

test_codered_banner

I restored the `wagtailadmin/base.forked.html` template removed in #226 to add the banner back to the CMS. However, I placed the banner within Wagtail's `content` div rather than between the sidebar and main content. Placing the banner in locations other than within that `content` div produced an unnecessary vertical scrollbar or other visual quirks.

If cleaner template code is preferred over a pleasant dev CMS nav, I could move the banner's `include` into `wagtailadmin/base.html` instead.

#### Documentation
The fix is documented in the `v0.23.0` release notes.

#### Tests
The new `coderedcms.tests.test_templates` module tests that the banner appears according to the `CODERED_BANNER` setting in both the public site and the CMS.


vsalvino:
Excellent, thanks for this and especially for including unit tests!

Now I remember why we got rid of base.forked.html... lots of duplication between wagtail, which changes with each wagtail version. If you can find a way to move the banner into base.html in a way that doesn't require re-implementing wagtail stuff, that would be ideal. If need be, we can add a bit of CSS in `codered-admin.css` to style the banner to help eliminate any visual quirks (e.g. position absolute or something)

James-Osgood:
I was able to move the banner into `base.html` with a bit of padding added to the main content container to make room for the banner (when it's there). It's nice to get that in there without wholesale copy-pasting the Wagtail template.

coderedcms/static/coderedcms/css/codered-admin.css
```diff
@@ -290,6 +290,14 @@ input[type='checkbox']::before, input[type='radio']::before {
     }
 }
 
+/* Display the optional banner above main content */
+
+.codered-banner {
+    position: absolute;
+}
+.codered-banner ~ .content-wrapper {
+    padding-top: calc(1.5em + 8px);
```
vsalvino, line 299: These seem like oddly specific magic numbers. Is there any reasoning for them? Would it perhaps make sense to set a fixed height of the banner instead, to make it easier to work with? I would be OK with such as change as the banner is purely utility and really only designed to hold one line of text / simple HTML.

James-Osgood: The banner's height is the text (`1.5em`) plus the bottom and top padding (`4px` x 2) hard-coded in the banner's inline `style` attribute. If that hard-coded padding ever changed, then this rule would need to also change.

Were you thinking of fixing the height in CSS (so also in `codered-front.scss`) rather than using the inline padding?

vsalvino (approved):

Issue #433, opened 2021-08-20
Image Block should default to original image, max-width 100% 


rcoldiron:
#### Is your feature request related to a problem? Please describe.
Images that use the default block can look stretched out and grainy at forced width of 100%. The automatic cropping also ruins some images. I am overriding this template lots of times to improve the look of images.

Also, maybe image uploads in general should default to this.

#### Describe the solution you'd like
Update template to keep image at its original proportions with max-width 100%. 

#### Additional context
If this is default, I won't be having to constantly recreate my template that overrides it. :)

PR #468, opened 2022-02-03
Issue433/imageblock

rcoldiron:
#### Description of change
Updated page cover images, image block, image link block, and hero block to use original image size instead of pre-cropping.
This way images aren't stretched or warped. 

Issue #433

### DIFF
--- a/coderedcms/templates/coderedcms/blocks/hero_block.html
+++ b/coderedcms/templates/coderedcms/blocks/hero_block.html
@@ -4,7 +4,7 @@
 <div class="container">
 {% endif %}
 
-{% image self.background_image max-2000x2000 as background_image %}
+{% image self.background_image original as background_image %}
 <div class="hero-bg {% if self.is_parallax %}parallax{% endif %} {% if self.tile_image %}tile{% endif %} {{self.settings.custom_css_class}}"
     style="{% if self.background_color %}background-color:{{self.background_color}};{% endif %}{% if background_image %}background-image:url({{background_image.url}});{% endif %}"
     {% if self.settings.custom_id %}id="{{self.settings.custom_id}}"{% endif %}>
--- a/coderedcms/templates/coderedcms/blocks/image_block.html
+++ b/coderedcms/templates/coderedcms/blocks/image_block.html
@@ -1,6 +1,6 @@
 {% load wagtailimages_tags %}
 
-{% image self.image max-1000x1000 as self_image %}
+{% image self.image original as self_image %}
 
 {% if format == 'amp' %}
 
--- a/coderedcms/templates/coderedcms/blocks/image_link_block.html
+++ b/coderedcms/templates/coderedcms/blocks/image_link_block.html
@@ -6,6 +6,6 @@
         data-ga-event-label='{{self.settings.ga_tracking_event_label|default:self.alt_text}}'
     {% endif %}
     title="{{ self.alt_text|safe }}" {% if self.settings.custom_id %}id="{{ self.settings.custom_id }}" {% endif %}>
-    {% image self.image max-1000x1000 as self_image %}
+    {% image self.image original as self_image %}
     <img src="{{self_image.url}}" class="img-fluid {{self.settings.custom_css_class}}" alt="{{self.alt_text|safe}}" />
 </a>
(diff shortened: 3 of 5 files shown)


coderedcms/templates/coderedcms/blocks/hero_block.html
```diff
@@ -4,7 +4,7 @@
 <div class="container">
 {% endif %}
 
-{% image self.background_image max-2000x2000 as background_image %}
+{% image self.background_image original as background_image %}
```
vsalvino, line 7: So in this case, `max-2000x2000` is not changing the proportions, but ensuring that the image is not greater than 2,000px in either direction (e.g. if someone uploads a giant 10MB 50,000px  50,000px image).

I'm not sure if we want to remove that safeguard?

coderedcms/templates/coderedcms/blocks/image_block.html
```diff
@@ -1,6 +1,6 @@
 {% load wagtailimages_tags %}
 
-{% image self.image max-1000x1000 as self_image %}
+{% image self.image original as self_image %}
```
vsalvino, line 3: Ditto on the max parameter. I think the problem that was causing you trouble before was the `w-100` class below.

Maybe we should use `max-2000x2000` as a safeguard, but replace `w-100` with `img-fluid` to ensure the image will scale, but not be forced to 100% width if it is smaller?

coderedcms/templates/coderedcms/blocks/image_link_block.html
```diff
@@ -6,6 +6,6 @@
         data-ga-event-label='{{self.settings.ga_tracking_event_label|default:self.alt_text}}'
     {% endif %}
     title="{{ self.alt_text|safe }}" {% if self.settings.custom_id %}id="{{ self.settings.custom_id }}" {% endif %}>
-    {% image self.image max-1000x1000 as self_image %}
+    {% image self.image original as self_image %}
```
vsalvino, line 9: Ditto on max comment.

coderedcms/templates/coderedcms/pages/article_page.html
```diff
@@ -5,7 +5,7 @@
 {% block content %}
     <article class="codered-article {%if self.cover_image %}has-img{% endif %}">
         {% if self.cover_image %}
-        {% image self.cover_image fill-2000x1000 as cover_image %}
+        {% image self.cover_image original as cover_image %}
```
vsalvino, line 8: In this case `fill-2000x1000` is going to scale/crop the image to force all blog cover images to be the same size.

I think this is probably the desired default behavior, but our template should make it easier to override. Maybe we should wrap these couple lines of image code in `{% block cover_image %}` or something, so that it can be overridden without needing to override everything?

coderedcms/templates/coderedcms/pages/web_page.html
```diff
@@ -3,7 +3,7 @@
 
 {% block content_pre_body %}
   {% if self.cover_image %}
-  {% image page.cover_image fill-2000x1000 as cover_image %}
+  {% image page.cover_image original as cover_image %}
```
vsalvino, line 6: Same comment as cover image on article_page

coderedcms/templates/coderedcms/blocks/hero_block.html
rcoldiron: Okay, reverting it back.

coderedcms/templates/coderedcms/pages/article_page.html
```diff
@@ -5,8 +5,10 @@
 {% block content %}
     <article class="codered-article {%if self.cover_image %}has-img{% endif %}">
         {% if self.cover_image %}
+        {% block cover_image %}
```
vsalvino, line 8: Blocks generally should go outside of if/else statements. Because if the `if` is not true, it will cause weird behavior.

coderedcms/templates/coderedcms/pages/article_page.html
```diff
@@ -5,8 +5,10 @@
...
         {% image self.cover_image fill-2000x1000 as cover_image %}
-        <img src="{{cover_image.url}}" class="w-100" />
+        <img src="{{cover_image.url}}" class="img-fluid" />
```
vsalvino, line 10: I think on the cover image we always want it to be w-100, right?

coderedcms/templates/coderedcms/pages/web_page.html
```diff
@@ -3,6 +3,7 @@
 
 {% block content_pre_body %}
   {% if self.cover_image %}
+  {% block cover_image %}
```
vsalvino, line 6: This template doesn't seem to need the block, because it already is within a `block content_pre_body`.

vsalvino (approved):

PR #483, opened 2022-05-12
Remove as much custom CSS as possible from the Wagtail admin

vsalvino:
* Remove as much of our custom CSS as possible, to reduce maintenance burden. This includes anything mildly opinionated, such as font styles, and anything that was used to fix previous accessibility issues in Wagtail which are no longer prevalent (e.g. light gray on white form inputs).
* Unfortunately most of the grisly CSS is still needed to make our nested streamfields usable. We should hopefully be able to remove nearly all of it with the UI changes coming in Wagtail 3.0.
* Fix showing the site logo in the wagtail admin for 2.16
* Remove our "powered by CodeRed" logo

Tested manually by spot-checking various areas of the Wagtail Admin.


jlchilders11:
The icons on our blocks seem to no longer be present. This may have been broken in a different PR, but we should probably fix it while updating the css.

vsalvino:
Seems like the icon breakage is related to the Wagtail 2.15/16 upgrade. Asked for more details here: https://github.com/wagtail/wagtail/issues/8533

vsalvino:
Logged a separate issue, since that bug is present in the dev branch: #484

FlipperPA:
@vsalvino I just have to say, this looks like it was **REALLY** fun to do! 🥇

vsalvino:
@FlipperPA the real fun is yet to come with Wagtail 3.0 (although it will be most fun if I get to delete the entire file)

PR #490, opened 2022-05-24
fix orderables, need to inherit Meta to work correctly

adremides:
#### Description of change
Make orderables save correctly the order. Due to the overwritting of Meta class, orderables didn't have the needed ordering variable. Here is the description of the [Orderable bug](https://github.com/wagtail/wagtail/issues/1749) in Wagtail repo.

#### Documentation
N/A

#### Tests
N/A

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -1042,7 +1042,7 @@ class CoderedEventIndexPage(CoderedWebPage):
 
 
 class CoderedEventOccurrence(Orderable, BaseOccurrence):
-    class Meta:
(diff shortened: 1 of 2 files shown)


vsalvino:
Thanks for the fix. I don't recall seeing any bug reports about this... is this bug present only on the most recent version, 0.23 (with Wagtail 2.16)? Can you provide a little background on the bug for the release notes? And similarly, should website owners also make the corresponding change in their customer orderable models?

vsalvino:
Thanks for the info. This looks good, we actually just confirmed this fixes a related bug! I think it might need a migration but we will add that after merging.

PR #501, opened 2022-06-24
Search all Wagtail Page models

### LINKED ISSUE #82: Database search backend is too bad
Wagtail's build-in database search backend does not provide good search results in coderedcms due to the fact that it can't search multiple fields on multiple different models, it can only search fixed fields on ALL models or multiple fields per-model. We have some workarounds in coderedcms for this (see [searching docs](https://docs.coderedcorp.com/cms/stable/features/searching.html)), but they are not ideal.

The elasticsearch backend is probably ideal, but adds an extra infrastructure component which is overkill for smaller sites. Similarly, there is a postgres backend but that does not support mysql or sqlite.

I recently discovered https://github.com/wagtail/wagtail-whoosh which would probably solve all of our problem, assuming it is actually portable between all database types. The one trick here is the need to run `update_index` to relfect changes, but we *might* be able to run that on a hook when a new page is published or when the cache is cleared. At any rate, if it does have to be run via cron job, I think cron is an "acceptable" piece of infrastructure moreso than requiring a specific database or elasticsearch.

Alternatively, I would love to write a mysql search backend that takes advantage of powerful full text indexing in mysql - but do not have enough knowledge about search backends to accomplish that in a timely manner. Regardless, I think wagtail-whoosh might be preferable to database-dependent backends.

jlchilders11:
#### Description of change
In Wagtail 2.16, they overhauled the search functionality to be able to properly search multiple different Page types. We had implemented this feature in a work around method locally in CRX, which is no longer necessary.

Fixes #82 

This change:
1. Removes the workaround to use Wagtail's search directly.
2. Maintains existing legacy behavior for page type filtering, but also allows filtering of non-CRX page types.
3. Deprecates several attributes from CRX Page models that are no longer necessary

#### Documentation
Updated search documentation to be in line with new functionality/deprecation.

#### Tests
A unit test for the search endpoint already exists, this cod change was made in line with it.

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -126,7 +126,7 @@ class CoderedPageMeta(PageBase):
     def search_db_include(self):
         warnings.warn(
             (
-                "This attribute of CoderedPageMeta has been depricated in favor of"
+                "This attribute of CoderedPageMeta has been deprecated in favor of"
                 " Wagtail's included search functionality."
             ),
             DeprecationWarning,
@@ -137,7 +137,7 @@ class CoderedPageMeta(PageBase):
     def search_db_boost(self):
         warnings.warn(
             (
-                "This attribute of CoderedPageMeta has been depricated in favor of"
+                "This attribute of CoderedPageMeta has been deprecated in favor of"
                 " Wagtail's included search functionality."
             ),
             DeprecationWarning,
@@ -148,7 +148,7 @@ class CoderedPageMeta(PageBase):
     def search_name(self):
         warnings.warn(
             (
-                "This attribute of CoderedPageMeta has been depricated in favor of"
+                "This attribute of CoderedPageMeta has been deprecated in favor of"
                 " Wagtail's included search functionality."
             ),
             DeprecationWarning,
@@ -159,7 +159,7 @@ class CoderedPageMeta(PageBase):
     def search_name_plural(self):
         warnings.warn(
             (
-                "This attribute of CoderedPageMeta has been depricated in favor of"
+                "This attribute of CoderedPageMeta has been deprecated in favor of"
                 " Wagtail's included search functionality."
             ),
             DeprecationWarning,
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -107,16 +107,8 @@ def get_page_models():
 class CoderedPageMeta(PageBase):
     def __init__(cls, name, bases, dct):
         super().__init__(name, bases, dct)
-        if 'search_db_include' not in dct:
-            cls._search_db_include = False
-        if 'search_db_boost' not in dct:
-            cls._search_db_boost = 0
         if 'search_filterable' not in dct:
             cls.search_filterable = False
-        if 'search_name' not in dct:
-            cls._search_name = cls._meta.verbose_name
-        if 'search_name_plural' not in dct:
-            cls._search_name_plural = cls._meta.verbose_name_plural
         if 'search_template' not in dct:
             cls.search_template = 'coderedcms/pages/search_result.html'
         if not cls._meta.abstract:
@@ -131,7 +123,7 @@ class CoderedPageMeta(PageBase):
             ),
             DeprecationWarning,
         )
-        return self._search_db_include
+        return False
 
     @property
     def search_db_boost(self):
@@ -142,7 +134,7 @@ class CoderedPageMeta(PageBase):
             ),
             DeprecationWarning,
         )
-        return self._search_db_boost
+        return 0
 
     @property
     def search_name(self):
@@ -153,7 +145,7 @@ class CoderedPageMeta(PageBase):
             ),
             DeprecationWarning,
         )
-        return self._search_name
+        return self._meta.verbose_name
 
     @property
     def search_name_plural(self):
@@ -164,7 +156,7 @@ class CoderedPageMeta(PageBase):
             ),
             DeprecationWarning,
         )
-        return self._search_name_plural
+        return self._meta.verbose_name_plural
 
 
 class CoderedTag(TaggedItemBase):
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -108,19 +108,63 @@ class CoderedPageMeta(PageBase):
     def __init__(cls, name, bases, dct):
         super().__init__(name, bases, dct)
         if 'search_db_include' not in dct:
-            cls.search_db_include = False
+            cls._search_db_include = False
         if 'search_db_boost' not in dct:
-            cls.search_db_boost = 0
+            cls._search_db_boost = 0
         if 'search_filterable' not in dct:
             cls.search_filterable = False
         if 'search_name' not in dct:
-            cls.search_name = cls._meta.verbose_name
+            cls._search_name = cls._meta.verbose_name
         if 'search_name_plural' not in dct:
-            cls.search_name_plural = cls._meta.verbose_name_plural
+            cls._search_name_plural = cls._meta.verbose_name_plural
         if 'search_template' not in dct:
             cls.search_template = 'coderedcms/pages/search_result.html'
         if not cls._meta.abstract:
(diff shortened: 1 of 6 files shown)


coderedcms/models/page_models.py
```diff
@@ -108,20 +108,64 @@ class CoderedPageMeta(PageBase):
     def __init__(cls, name, bases, dct):
         super().__init__(name, bases, dct)
         if 'search_db_include' not in dct:
-            cls.search_db_include = False
+            cls._search_db_include = False
```
vsalvino, line 111: Let's completely remove the `search_db_*` fields. They were a poor man's scoring, but wagtail search actually sorts by relevancy now.

coderedcms/models/page_models.py
```diff
@@ -108,20 +108,64 @@ class CoderedPageMeta(PageBase):
...
+            (
+                "This attribute of CoderedPageMeta has been depricated in favor of"
+                " Wagtail's included search functionality."
+            ),
+            DeprecationWarning,
+        )
+        return self._search_db_boost
+
+    @property
+    def search_name(self):
```
vsalvino, line 148: I think we will want to keep `search_name` and `search_name_plural`.

coderedcms/templatetags/coderedcms_tags.py
```diff
@@ -184,3 +184,8 @@ def map_to_bootstrap_alert(message_tag):
         return message_to_alert_dict[message_tag]
     except KeyError:
         return ''
+
+
+@register.filter
+def get_name_of_class(class_type):
+    return class_type.__class__.__name__
```
vsalvino, line 191: This should get the `search_name`, and as a fall back the `verbose_name`

coderedcms/views.py
```diff
@@ -39,49 +36,35 @@ def search(request):
         search_query = search_form.cleaned_data['s']
         search_model = search_form.cleaned_data['t']
 
-        # get all codered models
-        pagemodels = sorted(get_page_models(), key=lambda k: k.search_name)
-        # get filterable models
+        # get all page models
+        pagemodels = sorted(get_page_models(), key=lambda k: k.__name__)
```
vsalvino, line 40: Should use `search_name` here. Might need to refactor this to be an attribute on the class that gracefully handles a few levels of fallback.

coderedcms/views.py
```diff
@@ -39,49 +36,35 @@ def search(request):
...
+        # If any types have the search_filterable flag set, only show those
```
vsalvino, line 41: The filtering can continue to work as it used to. This is only used to show additional tabs in the UI. If it is empty, that's fine.

docs/features/searching.rst
```diff
@@ -14,53 +14,18 @@ The template can be overridden per model with the ``search_template`` attribute.
...
 
     class Product(CoderedWebPage):
         search_filterable = True
-        search_name = 'Product'
-        search_name_plural = 'Products'
 
 Would enable the following filter options on the search page: All Results, Blog, Products.
-
-Search fields
--------------
```
vsalvino, line 36: Keep this section, but add a versionremoved admonition below the heading indicating these were removed in 0.25, and sites should use the new search backend (link to wagtail docs).

coderedcms/views.py
jlchilders11: This cannot be an attribute on every page class, as this gets every page model, including wagtail pages.

coderedcms/views.py
jlchilders11: The purpose of this change was to allow for sorting based on Wagtail models as well, in the case that CRX is added after initial work is done, while still allowing users to hide undesired pages if necessary.

vsalvino: So search_filterable is not supposed to behave how your comment indicates (only show those). Instead it supposed to only present the option of filtering by those models in the template. It does not limit the main search to those models. Just presents the option to filter by them.

The variable name should probably be renamed "filterable_pagetypes" or something more descriptive rather than just "pagetypes"

docs/features/searching.rst
```diff
@@ -14,27 +14,28 @@ The template can be overridden per model with the ``search_template`` attribute.
...
-        search_name = 'Product'
-        search_name_plural = 'Products'
 
 Would enable the following filter options on the search page: All Results, Blog, Products.
 
+
 Search fields
 -------------
 
+**Deprecated in 0.25, instead the Wagtail search parameters should be used. https://docs.wagtail.org/en/stable/reference/contrib/searchpromotions.html#module-wagtail.contrib.search_promotions**
```
vsalvino, line 37: Use the rst directive for this notice: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

Might be a bit more readable to put the link in the "Wagtail search parameters" text

coderedcms/templates/coderedcms/pages/search.html
```diff
@@ -46,7 +46,7 @@ <h2>{% trans 'Search for' %} “{{form.s.value}}”</h2>
                 {% for pt in pagetypes %}
                 <li class="nav-item">
                     {% query_update qs_nop 't' pt.content_type.model as qs_t %}
-                    <a class="nav-link {% if form.t.value == pt.content_type.model %}active{% endif %}" href="?{{qs_t.urlencode}}">{{pt.search_name_plural}}</a>
+                    <a class="nav-link {% if form.t.value == pt.content_type.model %}active{% endif %}" href="?{{qs_t.urlencode}}">{{pt|get_name_of_class}}</a>
```
vsalvino, line 49: This label should show the plural name of the class.

coderedcms/templates/coderedcms/pages/search_result.html
```diff
@@ -1,9 +1,11 @@
+{% load coderedcms_tags %}
+
 <div class="d-flex align-items-center">
     <h4><a href="{{page.url}}">{{page.title}}</a></h4>
-    {% if page.content_type.model_class in pagetypes %}<small class="ml-3 badge badge-secondary">{{page.search_name}}</small>{% endif %}
+    {% if page.content_type.model_class in pagetypes %}<small class="ml-3 badge badge-secondary">{{page|get_name_of_class}}</small>{% endif %}
```
vsalvino, line 5: This label should show the singular name of the class.

docs/features/searching.rst
```diff
@@ -14,27 +14,28 @@ The template can be overridden per model with the ``search_template`` attribute.
 
 Search result filtering
 -----------------------
+By default, search results can be filtered by the page type to allow for a more specific search.
 
-To enable additional filtering by page type, add ``search_filterable = True`` to the page model.
-The ``search_name`` and ``search_name_plural`` fields are then used to display the labels for
```
vsalvino, line 19: Can probably restore most of this documentation about search_name

docs/features/searching.rst
```diff
@@ -32,9 +32,12 @@ For example, to enable search filtering by Blog or by Products in addition to Al
 
 Would enable the following filter options on the search page: All Results, Blog, Products.
 
+
 Search fields
 -------------
 
+.. deprecated:: 0.25, instead the `Wagtail search parameters <https://docs.wagtail.org/en/stable/reference/contrib/searchpromotions.html#module-wagtail.contrib.search_promotions>`_ should be used.
```
vsalvino, line 39: I believe after the version number, the text should be on a separate line and indented. E.g.
```
.. deprecated:: 0.25

   This is the message.
```

coderedcms/models/page_models.py
```diff
@@ -122,6 +118,28 @@ def __init__(cls, name, bases, dct):
         if not cls._meta.abstract:
             CODERED_PAGE_MODELS.append(cls)
 
+    @property
+    def search_db_include(self):
```
vsalvino, line 122: Do deprecation notices work if you continue to specify search_db_include on a model, the old way? Or would your attribute end up overriding this pre-defined one?

And the flip side, if you call `page.search_db_include`, did you confirm this warning actually fires?

vsalvino (approved):

PR #502, opened 2022-06-27
Implement Accordion Block

jlchilders11:
#### Description of change
This adds a new content block known as an Accordion Block. This implements the standard bootstrap accordion/collapse functionality in a reusable way: https://getbootstrap.com/docs/4.6/components/collapse/

#### Documentation
A new page explaining the fields and what this is was added under features/blocks/contentblocks/accordion.rst

#### Tests
This was run locally, the block was added and functionality tested.


coderedcms/blocks/__init__.py
```diff
@@ -82,6 +83,10 @@
 ]
 
 CONTENT_STREAMBLOCKS = HTML_STREAMBLOCKS + [
+    ('accordion', AccordionBlock([
```
vsalvino, line 86: Does the streamfield feel noticeably slower after this change? I know that every time we add another nested block it's an n-squared increase in page load time when rendering the streamfield (at least used to be, not sure in current version of wagtail?)

coderedcms/blocks/__init__.py
```diff
@@ -82,6 +83,10 @@
...
+        ('row', GridBlock(HTML_STREAMBLOCKS)),
```
vsalvino, line 87: Do we really want grids inside of accordions? I think this is adding way too many levels. Previously our max possible was 5 levels deep. This increases it to 7.

HeroBlock > GridBlock > ColumnBlock > AccordionBlock > GridBlock > ColumnBlock > Text.

coderedcms/templates/coderedcms/blocks/accordion_block.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'coderedcms/blocks/base_block.html' %}
+
+{% load wagtailcore_tags coderedcms_tags %}
+
+{% block block_render %}
+
+{% generate_random_id as modal_id %}
+<a
+  href="#"
```
vsalvino, line 9: I think accessibility standards might prefer this to be a button rather than an anchor, since it does not actually link to anywhere. Not sure if bootstrap4 supports this or not though. Just wanted to check.

docs/features/blocks/contentblocks/accordion.rst
```diff
@@ -0,0 +1,22 @@
+Modal Block
```
vsalvino, line 1: Title

coderedcms/blocks/__init__.py
jlchilders11: I didn't notice any particular slow down when testing locally.

coderedcms/blocks/content_blocks.py
```diff
@@ -12,6 +12,19 @@
 from .html_blocks import ButtonBlock
 
 
+class AccordionBlock(ButtonMixin, BaseBlock):
+    """
+    Wraps child comonents in a .collapse element, to allow them to be shown/hidden
+    """
+
+    carousel = SnippetChooserBlock('coderedcms.Accordion')
```
vsalvino, line 20: Probably should name this something more relevant.

coderedcms/templates/coderedcms/blocks/accordion_block.html
```diff
@@ -0,0 +1,19 @@
+{% extends 'coderedcms/blocks/base_block.html' %}
+
+{% load wagtailcore_tags coderedcms_tags %}
+
+{% block block_render %}
+
+{% generate_random_id as modal_id %}
```
vsalvino, line 7: Since we have a snippet here, we can use the snippet's ID rather than having to generate a random one. The generation is necessary for arbitrary blocks in the streamfield that don't have IDs. Something like `snippet-{id}` would work fine here.

coderedcms/blocks/content_blocks.py
```diff
@@ -12,6 +12,19 @@
 from .html_blocks import ButtonBlock
 
 
+class AccordionBlock(ButtonMixin, BaseBlock):
```
vsalvino, line 15: In this case, I would prefer to not use ButtonMixin, only because things like setting page, url, doc links are not relevant. Might be better off to inline a `title` field on this model and keep it simple.

coderedcms/templates/coderedcms/blocks/accordion_block.html
```diff
@@ -0,0 +1,19 @@
...
+<button
```
vsalvino, line 8: Let's make this template match the bootstrap docs example as closely as possible. Generally speaking the bootstrap examples are a baseline for our components.

https://getbootstrap.com/docs/4.6/components/collapse/#accordion-example

coderedcms/blocks/content_blocks.py
jlchilders11: In this case, all the button mixin does is add the three fields for the title, style, and size. Button Link is a different thing: 
image

coderedcms/models/snippet_models.py
```diff
@@ -343,6 +343,14 @@ def __str__(self):
         return self.name
 
 
+@register_snippet
+class Accordion(ReusableContent):
```
vsalvino, line 347: This should work exactly like the carousel... e.g. `Accordion` which contains many orderables of `AccordionItems` (for example).

I would also limit the accordion content to HTML_CONTENT, rather than the full nest of hero units, columns etc. It becomes unwieldy to edit/render.

coderedcms/templates/coderedcms/blocks/accordion_block.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'coderedcms/blocks/base_block.html' %}
+
+{% load wagtailcore_tags coderedcms_tags %}
+
+{% block block_render %}
+
+{% with a_id=self.accordion.id %}
+
+<button class="btn {{self.button_size}} {{self.button_style}}" type="button" data-toggle="collapse" data-target="#accordion-{{a_id}}" aria-expanded="false" aria-controls="accordion-{{a_id}}">{{self.button_title}}</button>
```
vsalvino, line 9: This template should look and function exactly like the bootstrap example: https://getbootstrap.com/docs/4.6/components/collapse/#accordion-example

image

coderedcms/blocks/content_blocks.py
vsalvino: Please take away the button mixin. This should function exactly like the carousel where they choose one Accordion snippet, and it renders the whole thing.

docs/features/blocks/contentblocks/accordion.rst
```diff
@@ -0,0 +1,22 @@
...
+===============
+
+Creates a collapsible section with a button to toggle.
+
+Field Reference
+---------------
+
+Fields and purposes:
+
+* **Button Title** - The text to show on the button. You can insert simple HTML
```
vsalvino, line 11: Docs should match fields on the model

coderedcms/models/snippet_models.py
```diff
@@ -349,6 +349,71 @@ def __str__(self):
...
+        related_name='accordion_panels',
+        verbose_name=_('Accordion'),
+    )
+
+    name = models.CharField(
+        max_length=255,
+        verbose_name=_('Name'),
+    )
+
+    content = CoderedStreamField(HTML_STREAMBLOCKS, blank=True)
```
vsalvino, line 395: Not sure why the unit tests _didn't_ throw a warning, but I'm pretty sure wagtail 3 now has a required argument `use_json_field=True`

https://docs.wagtail.org/en/stable/releases/3.0.html#use-json-field-argument-added-to-streamfield

PR #505, opened 2022-07-20
Fix ical file generation

### LINKED ISSUE #467: Importing an ICS file into "The New Outlook for Mac" gets stuck trying to sync to the living version of the event
Note: this only happens on "The New Outlook for Mac", not the previous version.

When you download an ICS file from an event page and upload it to Outlook, the event does not fully sync to the calendar. According to my research, this is default by Microsoft. 

- When a calendar is imported into Outlook via an ICS file, the event will not be synced to the Cloud.
- "Note: When you import an .ics file, your calendar doesn't refresh the imported events, even if the calendar's owner makes an update." - https://support.microsoft.com/en-gb/topic/cff1429c-5af6-41ec-a5b4-74f2c278e98c 
- Syncing allows for editing of an event. Because the event comes from someone else's calendar (ie the website where you downloaded the event), Outlook for Mac wants to sync to the calendar it came from but can't since the file is a static copy of the event. However, you can still delete an imported event from your own personal calendar.
- "### The difference between importing and subscribing to a calendar
When you import an .ics file, you get a snapshot of the events in the calendar at the time of import. Your calendar doesn't refresh the imported events automatically -- even if the calendar's owner makes an update. This is a good way to add events to your existing calendar that aren't going to change, like tide tables or phases of the moon.
Subscribing to an online calendar connects it to your calendar. Whenever the owner of a calendar you've subscribed to makes changes to the events, Outlook.com updates your calendar. This is a good way to track events that change frequently, like movie times or a school calendar." https://support.microsoft.com/en-gb/topic/cff1429c-5af6-41ec-a5b4-74f2c278e98c 
- Maybe we should allow for subscribing to the website's calendar or event to receive updates and sync instead of downloading an ICS file, or allow for both options. Subscribing still does not permit an outside user to edit someone else's event but it will refresh the event if info changes.

jlchilders11:
#### Description of change

Fixes #467 

As it stands, Outlook for Mac has a higher standard of compliance to the iCalendar standard, which our EventOccurencePage exports to. This higher standard was causing imported event for our .ics files to fail on calendar sync, and be deleted from the calendar. 

To correct this three things have been changed:

1. Additional fields have been added to the calendar (prodid and version), which are required for compliance.
2. Additional fields have been added to the event (dtstamp and uid) which are required for compliance.
3. The datetimes presented for dt_start and dt_end have been converted to UTC. The existing timezoned datetimes were being improperly formatted, which was what was actually breaking the imports.

#### Tests
An event was manually exported from an existing project to and .ics file, and imported into a freshly installed outlook for mac. This event was successfully added and editable.

#### Notes
While these changes correct the existing issue, the fact that the package used for handling iCalendar did not enforce compliance, and that its formatting broke with utc timezones make it a liability to use. In the future it should probably be removed for a simple formatted string/dict instead.

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -4,6 +4,8 @@ Base and abstract pages used in Wagtail CRX.
 
 import json
 import logging
+import uuid
+import pytz
 import os
 import warnings
 from datetime import date, datetime
@@ -845,14 +847,20 @@ class CoderedEventPage(CoderedWebPage, BaseEvent):
         ical_event.add('summary', self.title)
         # needs to get full page url, not just slug
         desc_str = _('Details')
+        ical_event.add('dtstamp', datetime.now())
         ical_event.add('description', f'{desc_str}: {self.full_url}')
+        ical_event.add('uid', uuid.uuid4())
         if self.address:
             ical_event.add('location', self.address)
 
         if dt_start:
+            # Convert to utc to remove timezone confusion
+            dt_start = dt_start.astimezone(pytz.utc)
             ical_event.add('dtstart', dt_start)
 
             if dt_end:
+                # Convert to utc to remove timezone confusion
+                dt_end = dt_end.astimezone(pytz.utc)
                 ical_event.add('dtend', dt_end)
 
             # Add a reminder alarm
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -5,7 +5,6 @@ Base and abstract pages used in Wagtail CRX.
 import json
 import logging
 import uuid
-import pytz
 import os
 import warnings
 from datetime import date, datetime
@@ -847,7 +846,7 @@ class CoderedEventPage(CoderedWebPage, BaseEvent):
         ical_event.add('summary', self.title)
         # needs to get full page url, not just slug
         desc_str = _('Details')
-        ical_event.add('dtstamp', datetime.now())
+        ical_event.add('dtstamp', timezone.now())
         ical_event.add('description', f'{desc_str}: {self.full_url}')
         ical_event.add('uid', uuid.uuid4())
         if self.address:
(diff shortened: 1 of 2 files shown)


coderedcms/models/page_models.py
```diff
@@ -4,6 +4,8 @@
 
 import json
 import logging
+import uuid
+import pytz
```
vsalvino, line 8: I believe pytz was removed as a dependency of Django in 4.0. I would recommend we avoid using it if possible. Not sure if wagtail still pulls it in or not.

coderedcms/models/page_models.py
```diff
@@ -845,14 +847,20 @@ def convert_to_ical_format(self, dt_start=None, dt_end=None, occurrence=None):
         ical_event.add('summary', self.title)
         # needs to get full page url, not just slug
         desc_str = _('Details')
+        ical_event.add('dtstamp', datetime.now())
```
vsalvino, line 850: Should this be `timezone.now()` instead? Not sure what dtstamp should contain, if it needs a naive or a tz-aware time.

coderedcms/models/page_models.py
```diff
@@ -845,14 +847,20 @@ def convert_to_ical_format(self, dt_start=None, dt_end=None, occurrence=None):
...
         ical_event.add('description', f'{desc_str}: {self.full_url}')
+        ical_event.add('uid', uuid.uuid4())
```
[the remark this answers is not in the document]
jlchilders11: So, looking into the spec, it seems that a uid should be the same for a given event, but also contain no identifiable information: https://icalendar.org/New-Properties-for-iCalendar-RFC-7986/5-3-uid-property.html Are we okay with generating one of these when creating an EventOccurence?

coderedcms/views.py
```diff
@@ -150,6 +150,8 @@ def event_generate_single_ical_for_event(request):
 
     # Generate the ical file.
     ical = Calendar()
+    ical.add('prodid', '-//Wagtail CRX//')
+    ical.add('version', '2.0')
```
vsalvino, line 154: These should be added to the other views as well.

vsalvino (approved):

PR #534, opened 2022-11-03
Wagtail4.1 backwards compatable with wagtail<4.1

htran-ubed:
#### Description of change
Made compatible with wagtail 4.1

#### Documentation
https://docs.wagtail.org/en/stable/releases/4.0.html
https://docs.wagtail.org/en/stable/releases/4.1.html

#### Tests
Unit tests needed

### DIFF
--- a/coderedcms/models/tests/test_navbars_and_footers.py
+++ b/coderedcms/models/tests/test_navbars_and_footers.py
@@ -1,5 +1,5 @@
 from django.test import Client
-from wagtail.test.utils import WagtailPageTests
+from wagtail.test.utils import WagtailPageTestCase
 from wagtail.models import Site
 
 from coderedcms.tests.testapp.models import WebPage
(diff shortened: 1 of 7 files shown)


vsalvino:
Thanks!!!

FYI - we do not need to maintain backwards compatibility with other versions of Wagtail. We pin the exact version. So you can just use the Wagtail 4.1 way of doing things. It should make the code simpler.

htran-ubed:
Alright, thanks. Let me submit the other version then.On 3 Nov 2022, at 17:04, Vince Salvino ***@***.***> wrote:﻿
Thanks!!!
FYI - we do not need to maintain backwards compatibility with other versions of Wagtail. We pin the exact version. So you can just use the Wagtail 4.1 way of doing things. It should make the code simpler.

vsalvino:
Thanks for your work on this PR. I have incorporated it into #556

PR #555, opened 2023-02-10
Fix EventPage previews

jlchilders11:

#### Description of change
Slightly modifies the logic for preview of Event pages. In the event that a preview of an event is created, and no future occurrences exist, we instead return the latest event.

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -1648,7 +1648,7 @@ class CoderedSubmissionRevision(SubmissionRevision, models.Model):
 
 
 class CoderedSessionFormSubmission(SessionFormSubmission):
-    
+
     INCOMPLETE = "incomplete"
     COMPLETE = "complete"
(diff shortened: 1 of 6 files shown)


coderedcms/models/page_models.py
```diff
@@ -807,20 +807,22 @@ def most_recent_occurrence(self):
...
             # Triggers when a preview is initiated on an
             # EventPage because it uses a FakeQuerySet object.
             # Here we manually compute the next_occurrence
             occurrences = [e.next_occurrence() for e in self.occurrences.all()]
-            if occurrences:
+            # If there are no more occurrences, we find the last one instead
+            if occurrences and None not in occurrences:
                 return sorted(occurrences, key=lambda tup: tup[0])[0]
 
+        aoc = []
```
vsalvino, line 820: This is the block of code that returns the last occurrence, if both attempts above to find the next occurrence fail, right? If so can you add a comment about it.

PR #557, opened 2023-02-13
Resolve Issues with event structured data not appearing on event pages

### LINKED ISSUE #552: Bug related to wagtail-seo
I tried implementing Structured Data for an EventPage using wagtail-seo. 

As far as i understand the CRX Codebase, the Structured Data is supposed to be processed and renderd by [Line 62 in the event_page.html](https://github.com/coderedcorp/coderedcms/blob/5c28e730acaf8aaceb7b12b943853c78e5f60944/coderedcms/templates/coderedcms/pages/event_page.html#L62) Template. 

But is  _{% block struct_seo_extra %}_ ever processed? It is defined in _wagtailseo/struct_data.html_ which gets included  in the _base Template_. But as far as i  know, Django does not process blocks in included files. At least this is my current guess, why I could not get any Structured Data to render.

According to the wagtail-seo documentation {{ self.my_struct_recipe_json }} should be put directly in the template. 
Doing so in CRX  raises an AttributeError ( 'tuple' object has no attribute 'start' ) in [Line 832 in models/page_models.py](https://github.com/coderedcorp/coderedcms/blob/5c28e730acaf8aaceb7b12b943853c78e5f60944/coderedcms/models/page_models.py#L832) 

I figure because self.most_recent_occurrence returns a Tuple (0 = start, 1=end, 2=Reference to the Model).
So 
"startDate": next_occ.start
should be
"startDate": next_occ[0],
or 
"startDate": next_occ[2].start,

Am i missing something crucial here? 
Or did i stumble on two bugs? 

Django==4.0.8
coderedcms==1.0.1
wagtail-seo==2.3.0

jlchilders11:

#### Description of change
Add custom block to `base.html` to allow for custom seo data not included by default in wagtail seo.
Cleans up logic in event page seo generation to fix errors.

Closes #552

### DIFF
--- a/coderedcms/models/page_models.py
+++ b/coderedcms/models/page_models.py
@@ -7,7 +7,7 @@ import logging
 import os
 import warnings
 from datetime import date, datetime
-from typing import Dict, List, Optional, TYPE_CHECKING, Union
+from typing import Dict, List, Optional, TYPE_CHECKING, Union, Tuple
 
 # This is a requirement for icalendar, even if django doesn't require it
 import pytz
@@ -798,7 +798,9 @@ class CoderedEventPage(CoderedWebPage, BaseEvent):
         return self.query_occurrences(num_of_instances_to_return=10)
 
     @property
-    def most_recent_occurrence(self):
+    def most_recent_occurrence(
+        self,
+    ) -> Tuple[datetime, datetime, BaseOccurrence]:
         """
         Gets the next upcoming, or last occurrence if the event has no more occurrences.
         """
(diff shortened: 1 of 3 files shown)


coderedcms/templates/coderedcms/pages/base.html
```diff
@@ -166,6 +166,7 @@
   {% block custom_scripts %}{% endblock %}
 
   {% include "wagtailseo/struct_data.html" %}
+  {% block struct_seo_extra %}{% endblock %}
```
vsalvino, line 169: Perhaps it would make sense to put `{% include "wagtailseo/struct_data.html" %}` inside a block, to facilitate easy overriding. That would be backwards compatible with existing templates since it is a new block that they were not aware of.

coderedcms/models/page_models.py
```diff
@@ -827,7 +827,7 @@ def most_recent_occurrence(self):
 
     @property
     def seo_struct_event_dict(self) -> dict:
-        next_occ = self.most_recent_occurrence
+        next_occ = self.most_recent_occurrence[2]
```
vsalvino, line 830: Can you add a comment as to what these magic number indexes in the tuple represent (perhaps even a type hint)?

PR #568, opened 2023-04-12
Story/567 related pages

jlchilders11:
#### Description of change
Implements three changes as part of issue 567:
    * Creates logic for "related pages", pages which are of a chosen type and share classifier terms
    * Creates the ability to generate "mini_views" of pages, which are used for previewing pages
    * Retrofit existing page chooser logic to instead use our new miniview logic

#### Documentation
TODO


coderedcms/models/page_models.py
```diff
@@ -533,6 +608,24 @@ def get_context(self, request, *args, **kwargs):
...
+        """
+        Render this page a small 'preview' component, suitable for inclusion on other
+        pages.
+
+        Set mini_view_template on this model to override what template is used, or
+        override coderedcms/templates/pages/mini_view_page.html to change all instances.
+        """
+        ctx = self.get_context(request=request, *args, **kwargs)
+        template_name = self.get_mini_render_template()
+        return render_to_string(template_name, ctx, request)
```
vsalvino, line 627: Not sure if I like this, it feels a bit backwards returning a string of rendered HTML..

I would instead opt to override `get_context` and include the queryset of related pages in the context. Then add a block in the base.html template for related content (similar to how we do for showing search results https://github.com/coderedcorp/coderedcms/blob/dev/coderedcms/templates/coderedcms/pages/search.html#L67).

coderedcms/settings.py
```diff
@@ -210,6 +210,10 @@ class _DefaultSettings:
             ),
         ],
         "pagepreviewblock": [
+            (
+                "coderedcms/blocks/miniview_page.html",
```
vsalvino, line 214: This can probably be omitted... The default (empty string) should be sufficient to trigger this template. In the future I'd like to remove these choices entirely since we have this new functionality.

coderedcms/templates/coderedcms/pages/form_mini_view
vsalvino, line 1: Empty file

coderedcms/models/page_models.py
```diff
@@ -455,6 +477,59 @@ def get_index_children(self):
...
+            r_qs = r_qs.annotate(p_ct=p_ct_q)
+            order_by.append("-p_ct")
+
+        # order by number of shared classifier terms
+        r_ct_q = models.Q(classifier_terms__in=self.classifier_terms.all())
+        r_qs = r_qs.annotate(r_ct=models.Count(r_ct_q))
+        order_by.append("-r_ct")
+
+        # if we include an index_order_by, order the set based on that
+        if self.index_order_by:
```
vsalvino, line 526: So, this will probably not work as expected. For example, if the current class is an ArticlePage (likely scenario)... it will not be using `index_order_by`, that is set on the parent ArticleIndexPage.

Perhaps a slightly more effective strategy... keep the related_querymodel, and if it is set run a query on all of those as you are currently doing. But if it is NOT SET, then simply get this page's parent, and call its `get_index_children()` (which already does all the ordering etc. for you).

coderedcms/models/page_models.py
jlchilders11: This is for rendering itself as a mini view. This is the generic ability to render itself as a mini preview, for part two of the ticket.

vsalvino:
I changed the templating a bit to make index, related, miniview, (and search) all work the same way using the miniview where possible. LMK what you think.

PR #579, opened 2023-05-27
Added closing figure tag to resolve issue #577.

### LINKED ISSUE #577: Missing closing figure tag in quote_block.html
#### Describe the bug
The missing `</figure>` tag in `quote_block.html` causes any custom CSS classes applied to quotes to affect all the following elements on the page.

#### Steps to reproduce
Steps to reproduce the behavior: (I was working on a Blog Article page at the time)
1. While editing a page, place a quote block above a text block with some content in both.
2. At the quote block, click on **Advanced Settings** and add `text-center` as a **Custom CSS Class**.
3. In the preview pane, watch the text in both blocks move to the center.

#### Expected behavior
Only the text in the quote block should center.

#### Additional context
I added the missing tag and it works as expected.
```
{% endif %}
</figure>
{% endblock %}
```

webvoyant:
#### Description of change
Added a closing `</figure>` tag in `quote_block.html` to fix the HTML structure and resolve #577.

#### Documentation
N/A

#### Tests
This change was tested manually in the preview pane while editing a page. If you put a quote block above a text block and add the `text-center` CSS class to the quote block, you'll see the behavior.

### DIFF
--- a/coderedcms/templates/coderedcms/blocks/quote_block.html
+++ b/coderedcms/templates/coderedcms/blocks/quote_block.html
@@ -6,4 +6,5 @@
 {% if self.author %}
 <figcaption class="blockquote-footer">{{self.author}}</figcaption>
 {% endif %}
+</figure>
 {% endblock %}


vsalvino:
Thank you!

PR #583, opened 2023-07-02
Added missing semicolons in film_strip_block template

cjkpl:
#### Description of change
Added missing semicolons in film_strip_block template. Without them, you cannot redefine both background and text colors.


vsalvino:
Excellent, thank you!

PR #587, opened 2023-07-12
Bug Fix mismatch of classifier name and slug max_length

jlchilders11:
#### Description of change
Change the max_length of the classifier slug field to match the classifier name field max_length.


coderedcms/models/snippet_models.py
```diff
@@ -138,9 +138,7 @@ class Meta:
         ordering = ["name"]
 
     slug = models.SlugField(
-        allow_unicode=True,
-        unique=True,
-        verbose_name=_("Slug"),
+        allow_unicode=True, unique=True, verbose_name=_("Slug"), max_length=255
```
vsalvino, line 141: Please keep exploded style.

PR #606, opened 2023-11-19
Adds Wagtail Formpages to the mix, too.

bahoo:
#### Description of change
I had a legacy form page which inherits from Wagtail's own `FormMixin`, rather than via CodeRed's models. This change simply includes those pages in the admin. My form submission results were inaccessible, otherwise.

#### Tests
1. Create a page model which inherits from Wagtail's forms, e.g.

```
from wagtail.contrib.forms.models import AbstractEmailForm
from wagtail.fields import RichTextField


class FormPage(AbstractEmailForm):
   intro = RichTextField(blank=True)
```
2. Make migrations, and migrate.
3. Create a new page for this `FormPage` in the Wagtail admin.
4. Verify that your newly created `FormPage` is now visible under the Forms tab in the admin.

### DIFF
--- a/coderedcms/wagtail_hooks.py
+++ b/coderedcms/wagtail_hooks.py
@@ -88,11 +88,12 @@ def crx_forms(user, editable_forms):
     and wagtail.contrib.forms.get_form_types()
     """
     from coderedcms.models import CoderedFormMixin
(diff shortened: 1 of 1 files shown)


vsalvino:
Good catch, thank you for contributing!

PR #610, opened 2023-11-27
Workaround for MySQL search backend bug

vsalvino:
Workaround for: https://github.com/wagtail/wagtail/issues/11273

Basically, using `.type()` + `.search()` + MySQL search backend + paginate results, creates a bug in which there are no search results.

* Run a slightly different query to avoid the bug which returns no search results.
* Also add a fallback search result template for vanilla Wagtail pages which do not inherit from CoderedPage.

### DIFF
--- a/coderedcms/views.py
+++ b/coderedcms/views.py
@@ -22,6 +22,8 @@ from django.views.decorators.http import require_POST
 from icalendar import Calendar
 from wagtail.admin import messages
 from wagtail.models import Page, get_page_models
+from wagtail.search.backends import get_search_backend
+from wagtail.search.backends.database.mysql.mysql import MySQLSearchBackend
 from coderedcms import utils
 from coderedcms.forms import SearchForm
 from coderedcms.models import CoderedPage, LayoutSettings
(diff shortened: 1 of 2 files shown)


coderedcms/views.py
```diff
@@ -61,7 +63,13 @@ def search(request):
                 model = ContentType.objects.get(
                     model=search_model
                 ).model_class()
-                results = results.type(model)
+                # Workaround for Wagtail MySQL search bug.
+                # See: https://github.com/wagtail/wagtail/issues/11273
+                backend = get_search_backend()
+                if type(backend) == MySQLSearchBackend:
+                    results = model.objects.live()
+                else:
```
jlchilders11, line 71: Is there a disadvantage to always doing `model.objects.live()`?

vsalvino: `.type(Model)` will query on Model, and also anything that inherits from Model. It is also the recommended approach in the docs. Doing `Model.objects.live()` will of course ONLY query on Model, not descendants.

jlchilders11 (approved):

PR #633, opened 2024-04-23
Include the absolute URL to the sitemap in the robots.txt file

FBruynbroeck:
#### Description of change
In the robots.txt file, the value of the sitemap field is a relative URL.
This should be changed to an absolute URL.
See protocol: https://sitemaps.org/protocol.html#submit_robots

For example, in the Google search console, this error message appears when it's a relative url:

> "Invalid sitemap URL" incomprehensible syntax

I think it would be interesting to have a default sitemap field in the right format using `root_url`.

What do you think of this change?


vsalvino:
Good catch, thanks!

PR #644, opened 2024-06-13
Fix Overflow on Carousel

hamishau:
Massive juicer of a pull request... Fixes horizontal scrollbar/overflow from appearing when carousel is in a fluid (full width) container. Applies to carousel in **Responsive Grid Row** and **Hero Unit**. Tested in FF and Chromium.

Screenshot from 2024-06-13 15-57-11

### DIFF
--- a/coderedcms/static/coderedcms/scss/_crx-bs-overrides.scss
+++ b/coderedcms/static/coderedcms/scss/_crx-bs-overrides.scss
@@ -15,7 +15,7 @@
 
 // Carousel
 .container-fluid .carousel {
-  margin: 0 -15px;
+  margin: 0 -12px;
 }
 
 .carousel .no-image {


coderedcms/static/coderedcms/scss/_crx-bs-overrides.scss
```diff
@@ -15,7 +15,7 @@
 
 // Carousel
 .container-fluid .carousel {
-  margin: 0 -15px;
+  margin: 0 -12px;
```
vsalvino, line 18: Have you tried removing this rule entirely? This was originally created on bootstrap 4, so I'm not sure if it's still even needed.

hamishau: Yes I have - it is required and is not relevant to Bootstrap 4/5, it is a CRX override. Without the negative margin, the "full width" carousel has horizontal whitespace. The negative margin accounts for --bs-gutter-x in Bootstrap fluid containers to calculate left and right padding.

-15px of margin creates horizontal overflow.
-12px of margin removes overflow and makes carousel full width in fluid containers.
- Removing rule entirely creates horizontal whitespace, see attached.

image

vsalvino:
Thanks... this looks good. Just one step is needed before I can merge this. Need to compile the css from the sass, then commit the result: https://docs.coderedcorp.com/wagtail-crx/contributing/index.html#css-development

PR #664, opened 2024-11-25
Provide time_zone in event index page context

jlchilders11:
Adds `time_zone` to the `EventIndexPage` context, since the template uses that variable.


vsalvino:
Looking at the calendar template, `time_zone` is already defined. I don't see it being used on event_index_page.html directly. Was this a bug in a different version maybe? https://github.com/coderedcorp/coderedcms/blob/main/coderedcms/templates/coderedcms/includes/ical/calendar.html

vsalvino:
Closing this was fixed in a prior version. Feel free to re-open if it becomes an issue.

PR #666, opened 2024-12-10
Update list submission template to match new wagtail update template

jlchilders11:

#### Description of change
The `wagtailforms/list_submissions.html` template is customized to use/respect protected media. This was changed in Wagtail 6.1, but was not updated to match in the CRX, causing pagination to fail to appear.

### DIFF
--- a/coderedcms/templates/wagtailforms/list_submissions.html
+++ b/coderedcms/templates/wagtailforms/list_submissions.html
@@ -1,39 +1,49 @@
-{% load i18n coderedcms_tags %}
-
-<div class="overflow">
-    <table class="listing">
-        <col />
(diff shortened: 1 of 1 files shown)


vsalvino:
Considering our 4.0 supports any version of Wagtail 6.x, I don't think we would want to backport it to that version since there's a chance it could reverse-break on wagtail 6.0. Thoughts?

We will definitely want it in 4.1 though.

jlchilders11:
It seems our options are either we drop support for 6.0 in 4.1, or we add a second template and a version check that provides the template based on wagtail version. Or honestly both, if we want to support this particular overlap.

PR #668, opened 2024-12-27
Fix Card Footer

hamishau:
Improved the default card template to hide the card-footer div if a link is not added. Current shows an empty section which shouldn't be visible if there is no link:

image

Improved version hides the card footer if no link has been specified in the Wagtail editor:

image

### DIFF
--- a/coderedcms/templates/coderedcms/blocks/card_foot.html
+++ b/coderedcms/templates/coderedcms/blocks/card_foot.html
@@ -9,9 +9,9 @@
     {% if self.subtitle %}<h6 class="card-subtitle mb-2 text-muted">{{self.subtitle}}</h6>{% endif %}
     <div class="card-text">{{self.description}}</div>
   </div>
-  <div class="card-footer">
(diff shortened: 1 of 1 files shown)


coderedcms/templates/coderedcms/blocks/card_foot.html
```diff
@@ -9,9 +9,9 @@
     {% if self.subtitle %}<h6 class="card-subtitle mb-2 text-muted">{{self.subtitle}}</h6>{% endif %}
     <div class="card-text">{{self.description}}</div>
   </div>
-  <div class="card-footer">
-    {% for button in self.links %}
-    {% include_block button %}
-    {% endfor %}
-  </div>
+  {% for button in self.links %}
```
vsalvino, line 12: I think it would be best to leave the for-loop inside the div, and replace this with an `{% if self.links %}` check. 

For example, if there are two buttons, they should render side by side. This change would make them render in two separate footers.

Issue #673, opened 2025-01-23
How-To-Guide > Customize Navbar and Footer


DrBix:
Hi all, I am still learning and I have now reached this part of the Tutorial in the documentation, and I have 2 issues:

(1)  ...The logo and style of the navbar can be customized under Settings > Layout  <---  my installation doesn't appear to have 'Layout' under Settings. Did I miss something? Has 'Layout' moved or been removed?

(2) ...You can now create more than one navbar menu or footer and choose which ones to render on your site. In Settings > CRX Settings, select your navbars in Site Navbars.  <---- Under my CRX Settings I don't see any option to select a navbar for Site Navbars.  I had created 2 Navbars and the links from both navbars appear on my site. How / where do I specify which navbar to display for my website?

Any help would very much be appreciated.

Cheers!

vsalvino:
Thanks for reporting these issues. The tutorial is probably a little bit out of date due to some recent changes in CRX.

1) "Layout" and most other settings were moved to "CRX Settings". 

2) If you started with the `pro` template, or have `CRX_DISABLE_NAVBAR = True` in your django settings, the navbar settings will be hidden. Please check if you see that setting.

vsalvino:
If you'd like to make a pull request to fix the Layout vs CRX Settings in the docs, that would be helpful :)

PR #684, opened 2025-02-20
Update custom Navbar/Footer models & templatetags to work in multi-site

### LINKED ISSUE #683: Make the Custom Navbar/Footer in Pro Template work in Multi-Site Installs
#### Is your feature request related to a problem? Please describe.

We receently implemented WagtailCRX, and noticed that wagtail's Navbar and Footer models in the basic template work in multi-site installations, but the custom Navbar and Footer models in the pro template do not (though you could fall back to wagtail's default models with `CRX_DISABLE_NAVBAR = False` and `CRX_DISABLE_FOOTER = False`).

#### Describe the solution you'd like

We implemented the following in our site, which allows us to use custom Navbar(s) and Footer(s) in our multi-site install. Because nulls are allowed, it won't cause errors on existing installs. In existing installs, though, the template tags as implemented below would cause the components to not be rendered (existing model instances would not associated with any site initially). To resolve this, we could:

- Modify the template tags to check for a setting before implementing their site checks so it won't disappear the navbars and footers in existing projects (falling back to current functionality if `CRX_USE_SITE_FIELD=False` for example)
- Or, just set the template tags to fall back to showing all Navbar(s) and Footer(s) if there are no site-specific results

Would a PR to update the pro template (and add relevant tests & updates to docs) be of interest? Any additional thoughts/recommendations?

This proposal would also resolve #673 

---

models.py

```python
"""
Create or customize your page models here.
"""

from coderedcms.blocks import (
    HTML_STREAMBLOCKS,
    LAYOUT_STREAMBLOCKS,
    BaseBlock,
    BaseLinkBlock,
    LinkStructValue,
)
from coderedcms.forms import CoderedFormField
from coderedcms.models import (
    CoderedArticleIndexPage,
    CoderedArticlePage,
    CoderedEmail,
    CoderedEventIndexPage,
    CoderedEventOccurrence,
    CoderedEventPage,
    CoderedFormPage,
    CoderedLocationIndexPage,
    CoderedLocationPage,
    CoderedWebPage,
)
from django.db import models
from modelcluster.fields import ParentalKey
from wagtail import blocks
from wagtail.admin.panels import FieldPanel
from wagtail.fields import StreamField
from wagtail.snippets.models import register_snippet


# Other models...

@register_snippet
class Navbar(models.Model):
    """Custom navigation bar / menu."""

    class Meta:
        """Meta class for Navbar."""
        verbose_name = "Navigation Bar"

    name = models.CharField(
        max_length=255,
    )
    content = StreamField(
        [
            ("link", NavbarLinkBlock()),
            ("dropdown", NavbarDropdownBlock()),
        ],
        use_json_field=True,
    )

    site = models.ForeignKey(  # <-- New Field
        "wagtailcore.Site",
        on_delete=models.CASCADE,
        related_name="navbars",
        null=True,
        blank=True,
    )

    panels = [
        FieldPanel("name"),
        FieldPanel("site"),  # <-- New panel item
        FieldPanel("content"),
    ]

    def __str__(self) -> str:
        return self.name


@register_snippet
class Footer(models.Model):
    """Custom footer for bottom of pages on the site."""

    class Meta:
        """Meta class for Footer."""
        verbose_name = "Footer"

    name = models.CharField(
        max_length=255,
    )
    content = StreamField(
        LAYOUT_STREAMBLOCKS,
        verbose_name="Content",
        blank=True,
        use_json_field=True,
    )

    site = models.ForeignKey(  # <-- New Field
        "wagtailcore.Site",
        on_delete=models.CASCADE,
        related_name="footers",
        null=True,
        blank=True,
    )

    panels = [
        FieldPanel("name"),
        FieldPanel("site"),  # <-- New panel item
        FieldPanel("content"),
    ]

    def __str__(self) -> str:
        return self.name
```

Updated template tags render the navbar(s) and footer(s) for the current site:

website_tags.py

```python
"""Custom template tags for the website app."""
from django import template
from wagtail.models import Site
from website.models import Footer, Navbar

register = template.Library()


@register.simple_tag(takes_context=True)
def get_website_navbars(context):
    """Get the navbars for the current site.

    Args:
        context: The template context which contains the current request

    Returns:
        QuerySet: Navbar queryset filtered by the current site
    """
    try:
        # Get the current request from context
        request = context['request']
        # Get the current site from the request
        current_site = Site.find_for_request(request)
        # Return navbars associated with the current site
        return Navbar.objects.filter(site=current_site)
    except (KeyError, AttributeError):
        # Fallback to returning all navbars if we can't determine the current site
        return Navbar.objects.all()


@register.simple_tag(takes_context=True)
def get_website_footers(context):
    """Get the footers for the current site.

    Args:
        context: The template context which contains the current request

    Returns:
        QuerySet: Footer queryset filtered by the current site
    """
    try:
        # Get the current request from context
        request = context['request']
        # Get the current site from the request
        current_site = Site.find_for_request(request)
        # Return footers associated with the current site
        return Footer.objects.filter(site=current_site)
    except (KeyError, AttributeError):
        # Fallback to returning all footers if we can't determine the current site
        return Footer.objects.all()
```

jacklinke:
#### Description of change
Added a `site` FK field to the custom `Navbar` and `Footer` models in the pro template; added the new migration for these models; and updated the website templatetags in the pro template to work with the newly updated models.

Resolves #683 & #673 

#### Documentation
Updated the docs the discuss how the snippets in the basic and pro templates work with multi-site installs in different ways.

#### Tests
Did *not* add tests, because there are no existing tests for the pro template website app (unless I missed something). Without an existing test suite for the website app in which to add my tests, I didn't want to make significant additions without asking the coderedcorp folks how/if I should approach this.


coderedcms/project_template/pro/project_name/settings/base.py
```diff
@@ -194,3 +194,7 @@
 # custom implementation.
 CRX_DISABLE_NAVBAR = True
 CRX_DISABLE_FOOTER = True
+
+# Default to True to maintain existing behavior for existing installations.
+CRX_NO_SITE_NAVBAR_FALLBACK = True
```
vsalvino, line 199: These settings are not necessary. The project_template is only applied when someone is creating a new site via `coderedcms start`, so backwards compatibility is not necessary. Please remove them to simplify it.

coderedcms/project_template/pro/website/migrations/0003_footer_site_navbar_site.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 5.1.6 on 2025-02-17 15:16
```
vsalvino, line 1: We have a special process for generating migrations in the project_template.
1. Install the oldest supported version of Django (currently 4.2).
2. Make a fresh testproject with `coderedcms start --template pro testproject`.
3. In the new testproject, delete all the migrations in `website/migrations/`
3. Run `python manage.py makemigrations`
4. Copy `website/migrations/0001_initial.py` back into `project_template/pro/website/migrations/`

This process ensures that new projects get a "clean" starting point.

vsalvino:
Nice work, thanks! We can remove that fallback code as it is not necessary and should simplify it a bit.

Tests aren't necessary here, as long as you manually tested it. The pipeline will generate a pro project and run the full battery of tests using it.

jacklinke:
As requested.

Also, apologies for committing directly to main vice a new branch. I'm clearly a bit out of practice 😬

coderedcms/project_template/pro/website/models.py
```diff
@@ -221,9 +221,17 @@ class Meta:
         ],
         use_json_field=True,
     )
+    site = models.ForeignKey(
+        "wagtailcore.Site",
+        on_delete=models.CASCADE,
```
vsalvino, line 226: I think SET_NULL would be preferable, in the event that you wanted to switch this navbar to a different site and deleted the site first. Similar to how the root page works.

coderedcms/project_template/pro/website/models.py
```diff
@@ -249,9 +257,17 @@ class Meta:
         blank=True,
         use_json_field=True,
     )
+    site = models.ForeignKey(
+        "wagtailcore.Site",
+        on_delete=models.CASCADE,
```
vsalvino, line 262: Ditto SET_NULL

vsalvino:
Looking good, left a few more comments. I think it will be ready to merge after that.

Issue #685, opened 2025-03-02
Tutorial: using $card-height: 100% causes article list page to have a lot of whitespace later on.


cericoda:
I've just been following the excellent tutorial and ran into a hitch.

Following the tutorial, in Part 5, I chose the second option, and set $card-height: 100% in _variables.scss

https://github.com/coderedcorp/coderedcms/blob/39faf4fdcd703d6c2f4d3b09dce0ddc8699a5ec0/docs/getting_started/tutorial03.rst?plain=1#L198C35-L198C94

When I got to part 5, instead of the image presented in 
https://docs.coderedcorp.com/wagtail-crx/_images/landing_page_preview.jpeg 

I got a lot of whitespace and the article list overflowing the footer.

PR #687, opened 2025-03-07
docs fix - prevent future issue on Blog list #685

cericoda:
I've removed optional guidance to change to _variables.css since this causes issues later in the tutorial.


vsalvino:
Thanks!

PR #717, opened 2026-07-17
Fix 404 viewing form submissions in admin on Wagtail 7.x (#710)

### LINKED ISSUE #710: Form submissions return 404 with Wagtail 7.x due to FormMixin isinstance check
As @vsalvino and the team know, I am not a programmer and tend to find my way by trial and error (mostly error). So, like many people lately, I've been using Claude Code to assist with web development. The error description below was created by Claude in response to an error I am seeing. I apologize if it's inappropriate to post an issue generated by an LLM (it does look like a pretty good issue description, though...).

## Description

Form submissions are inaccessible in the Wagtail admin (404 error) when using CodeRed CMS 6.0 with Wagtail 7.x. The Forms listing page shows forms correctly, but clicking to view submissions returns a 404.

## Steps to Reproduce

1. Install CodeRed CMS 6.0.0 with Wagtail 7.1.x
2. Create a FormPage with some form fields
3. Submit the form a few times to create submissions
4. Go to Wagtail admin → Forms
5. Click on the form page to view submissions
6. **Result**: 404 error at `/admin/forms/submissions/<page_id>/`

## Root Cause

In May 2025, Wagtail added an `isinstance` check in [`wagtail/contrib/forms/views.py`](https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/forms/views.py#L37) (commit [efd5d49](https://github.com/wagtail/wagtail/commit/efd5d49438d90df5c73981983e873d31873edbde)):

```python
def get_submissions_list_view(request, *args, **kwargs):
    page_id = kwargs.get("page_id")
    form_page = get_object_or_404(Page, id=page_id).specific
    if not isinstance(form_page, FormMixin):
        raise Http404  # <-- This causes the 404
    return form_page.serve_submissions_list_view(request, *args, **kwargs)
```

The problem is that `CoderedFormMixin` does **not** inherit from `wagtail.contrib.forms.models.FormMixin`:

```python
>>> from wagtail.contrib.forms.models import FormMixin
>>> from coderedcms.models import CoderedFormPage, CoderedFormMixin
>>> issubclass(CoderedFormPage, FormMixin)
False
>>> issubclass(CoderedFormMixin, FormMixin)
False
```

## Suggested Fix

`CoderedFormMixin` should inherit from `wagtail.contrib.forms.models.FormMixin` to pass the isinstance check. This may require reconciling any method/attribute conflicts between the two mixins.

Alternatively, CodeRed CMS could register its own URL route for form submissions that bypasses Wagtail's check.

## Environment

- **CodeRed CMS**: 6.0.0
- **Wagtail**: 7.1.1
- **Python**: 3.14
- **Django**: 5.1.x

## Related

This likely also affects `wagtail-flexible-forms` since `StreamFormMixin` similarly does not inherit from Wagtail's `FormMixin`.

bradrice:
Fixes #710

## Problem

On Wagtail 7.x, viewing form submissions in the Wagtail admin returns a **404**. The Forms index lists the forms, but clicking through to a form's submissions (`/admin/forms/submissions/<id>/`) fails.

Wagtail 7.x [hardened its admin submissions view](https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/forms/views.py) to guard on the form type and use the new API:

```python
form_page = get_object_or_404(Page, id=page_id).specific
if not isinstance(form_page, FormMixin):   # CodeRed pages fail this
    raise Http404
return form_page.serve_submissions_list_view(request, ...)
# ...and SubmissionsListView.get_base_queryset() now calls form_page.get_submissions()
```

CodeRed's `CoderedFormMixin` / form pages do **not** inherit Wagtail's `wagtail.contrib.forms.models.FormMixin`, so both the `isinstance` guard and the `get_submissions()` call fail.

## Fix

Have `CoderedFormPage` and `CoderedStreamFormPage` inherit Wagtail's `FormMixin` **as the last base**:

```python
class CoderedFormPage(CoderedFormMixin, CoderedWebPage, FormMixin): ...
class CoderedStreamFormPage(CoderedFormMixin, CoderedStreamFormMixin, CoderedWebPage, FormMixin): ...
```

Why last, and why on the page classes rather than `CoderedFormMixin`:

- `FormMixin` is a **fieldless plain mixin** → **no migration** (`makemigrations --check` reports *No changes detected*).
- Placing it **last** in the MRO means it only supplies the two methods CodeRed lacks (`get_submissions()`, `get_submissions_list_view_class()`) and satisfies the `isinstance` check — it never shadows CodeRed's own methods.
- Putting it on `CoderedFormMixin` instead would insert `FormMixin` **ahead of** `wagtail_flexible_forms.StreamFormMixin` in `CoderedStreamFormPage`'s MRO, so `FormMixin.get_form_fields()` (which does `self.form_fields.all()`) would shadow the StreamField handling and break advanced forms (`'StreamValue' object has no attribute 'all'`). Inheriting on the page classes, last, avoids that.

Method resolution after the change:

| Page | `isinstance(.., FormMixin)` | `get_data_fields` | `get_submissions` |
|---|---|---|---|
| `CoderedFormPage` | ✅ | `CoderedFormPage` (unchanged) | `FormMixin` |
| `CoderedStreamFormPage` | ✅ | `StreamFormMixin` (unchanged) | `FormMixin` |

## Tests

Adds `coderedcms/tests/test_forms.py` covering the admin forms index and the submissions view for **both** the standard and advanced (StreamField) form pages. These tests **fail without the fix** (both submission views 404) and pass with it.

- Full suite: **72 passed**
- `ruff check` / `ruff format --check`: clean
- `codespell`: clean
- `makemigrations --check`: No changes detected

## Docs

Adds `docs/releases/v6.0.1.rst` (backwards-compatible bug fix → patch) and links it in the releases index.


bradrice:
Notes on the first CI run:

- ✅ **Unit tests pass on all matrix jobs** (py3.10–3.13, basic + pro templates).
- **Linters / Spelling** failed on a *pre-existing* codespell error in `docs/getting_started/tutorial05.rst:106` ("pre-selected"), which is present on `main` and unrelated to this change (codespell 2.4.3 now flags it). I've included a one-line commit fixing that word so the linters job passes — happy to drop it into a separate PR if you'd prefer to keep this one strictly scoped.
- **Code Coverage** failed in `compare-codecov.ps1`. This PR only *adds* code (a regression test covering the admin submissions path), so it shouldn't reduce coverage; this looks like the comparison flakiness the script itself notes, possibly amplified by the fork-PR baseline. Please re-run if needed.

The functional change is unaffected by either.

bradrice:
Update after the latest run (build 5158): unit tests, linters (spelling now fixed), and docs all pass. ✅

The only remaining red check is **Code Coverage**, and the log confirms it's not a coverage regression — `compare-codecov.ps1` fails on the `Invoke-WebRequest` call that fetches the `main` baseline from the Azure DevOps API (it never reaches the comparison). That looks like a fork-PR API-access limitation rather than anything in this change. A maintainer re-run from within the repo (or a manual override) should clear it.

bradrice:
Gentle bump — this has been open since July, with CI green apart from the known `compare-codecov.ps1` baseline fetch noted above.

Since filing, I've applied this fix as a per-project override across six production sites running coderedcms 6.0.0 / Wagtail 7.4.2. That surfaced an additional symptom of the same root cause which isn't in the description above, and which I think strengthens the case for the fix:

`wagtail.contrib.forms.utils.get_form_types()` filters on the same `FormMixin`, so it returns `[]` for CodeRed projects. The result is that the **Forms menu does not render in the admin at all** — the failure presents as a *missing menu* rather than a 404, which is substantially harder to diagnose. On a site with no form pages yet, it's entirely latent: nothing looks wrong until someone adds the first form and can't find where submissions live.

Measured before and after on four separate projects, identical on each:

```
before:  issubclass(FormPage, FormMixin) False | get_submissions absent | get_form_types() []
after:   True                                  | present               | ['Website | Form']
```

`makemigrations --check` reports *No changes detected*, and `get_form_fields()` still resolves to `CoderedFormPage`, so CodeRed's own behavior is unchanged.

Happy to rebase onto current `main`, or to split the codespell fix into a separate PR if that makes review easier.

